PR #28, opened 2017-02-20
add fmi link with raa-nr as qualifier

Vesihiisi:
Per discussions in: https://www.wikidata.org/wiki/Property_talk:P1262

Which means the RAÄ-number has to be implemented as a qualifier to Cultural heritage database in Sweden (P1260)

Task: https://phabricator.wikimedia.org/T158060


lokal-profil:
Which task does this correspond to?

Vesihiisi:
This is T158060.

importer/SeFornminSv.py
```diff
@@ -48,10 +48,18 @@ def set_descriptions(self):
 
     def set_raa(self):
         """
-        All items have a 'raa-nr'.
+        All items have a 'raa-nr' and an 'id'.
+        The latter is used to link to the
```
lokal-profil: It would be nice to have a blank line between the first line of the documentation and the rest. This is also in accordance with PEP257

PR #31, opened 2017-03-01
Upload live

Vesihiisi:
Uploads go live instead of to the Wikidata sandbox. Includes creating new items when necessary.

Task: https://phabricator.wikimedia.org/T159328


importer/Uploader.py
```diff
@@ -246,17 +245,18 @@ def upload(self):
     def set_wd_item(self):
         item_exists = True if self.data["wd-item"] is not None else False
         if item_exists:
-            # item_q = self.data["wd-item"]
-            self.wd_item = self.wdstuff.QtoItemPage(self.TEST_ITEM)
-            print(self.wd_item)
+            item_q = self.data["wd-item"]
```
lokal-profil: I would probably keep the test item options. Maybe something like:
```
if self.test:
    self.wd_item = self.wdstuff.QtoItemPage(self.TEST_ITEM)
elif item_exists:
    item_q = self.data["wd-item"]
    self.wd_item = self.wdstuff.QtoItemPage(item_q)
else:
    self.wd_item = self.create_new_item(self.log)
```
That way you can toggle testing on/off (which is useful for later datasets) without having to update deep parts of the code. You could even set it as a command line argument.

lokal-profil: Unless you now use a different testing strategy, of course.

Vesihiisi: I have addressed this now; just using the --upload argument on its own will use the Sandbox mode, in order to go `live` you actually have to add that as an argument. That way you have to explicitly request the live mode, which prevents triggering it by accident. I have also updated the README.

importer/Uploader.py
```diff
@@ -243,19 +243,33 @@ def upload(self):
...
+        """
+        Determine WD item to manipulate.
+
+        In live mode, if data object has associated WD item,
+        edit it. Otherwise, create a new WD item.
+        In sandbox mode, all edits are done on the WD Sandbox item.
+        """
+        if self.live:
+            item_exists = True if self.data["wd-item"] is not None else False
+            if item_exists:
```
lokal-profil: Any reason to not just use `if self.data["wd-item"] is not None` directly?

Are there situations when `if self.data["wd-item"]` is false but `self.data["wd-item"] is None` is also false?

importer/Uploader.py
lokal-profil: That seems like a good solution to me =)

importer/Uploader.py
Vesihiisi: The official reason is probably "I wasn't sure what would happen if I left that out and also I wanted to be explicit to myself"

Vesihiisi: Ok, the more I look at this the less sense it makes. I guess I thought a shorter "if" row would look better.

Vesihiisi: I have simplified this part.

PR #32, opened 2017-03-02
Unmatched report

Vesihiisi:
Create a json report for values which the bot expects to handle but could not parse/match for a particular instance.

The report is generated in the data processing stage. If an upload is included, and the particular data item does not have an associated Qid -- meaning the Uploader creates a new Wikidata item -- the Qid of this new item is requested from the Uploader and inserted into the report in the uploading stage. At the end of the process, a cumulative report for the batch is saved into a file.

To begin with, reporting functionality has been added to SeArbetslSv.py

A report contains:
- item Qid
- name of problematic field
- raw data of problematic field
- URL of data item in the WLM API

Example report -- in live mode, the empty Q fields would be filled with Qid's of newly created WD items: https://phabricator.wikimedia.org/P5004.

Task: https://phabricator.wikimedia.org/T159325


importer/importer.py
```diff
@@ -186,8 +188,19 @@ def get_items(connection,
...
+                this means the Uploader has now created a new Item
+                for it -- insert that id into the problem report.
+                """
+                problem_report["Q"] = uploader.wd_item_q
             uploader.upload()
             print("--------------------------------------------------")
+        problem_reports.append(problem_report)
+    non_empty_reports = utils.remove_empty_dicts_from_list(problem_reports)
+    utils.json_to_file(
+        "report_" + specific_table_name + ".txt", non_empty_reports)
```
lokal-profil: Should this be .json?

importer/SeArbetslSv.py
```diff
@@ -81,6 +86,7 @@ def set_type(self):
                 for special in special_type:
                     self.add_statement("is", special, refs=[ref])
             except IndexError:
+                self.add_to_report("typ", self.typ)
                 return
```
lokal-profil: can drop this `return`

importer/SeArbetslSv.py
```diff
@@ -100,6 +108,7 @@ def set_location(self):
                 ref = self.wlm_source
                 self.add_statement("location", location, refs=[ref])
             except IndexError:
+                self.add_to_report("ort", self.ort)
                 return
```
lokal-profil: Can drop this `return`

importer/importer.py

importer/SeArbetslSv.py
Vesihiisi: Done.

importer/SeArbetslSv.py
Vesihiisi: Done.

PR #33, opened 2017-03-03
use add_description() from wdstuff

Vesihiisi:
Use the description adding functionality in [Wikidatastuff](https://github.com/lokal-profil/wikidata-stuff) instead of the current home-assembled one, so that it's consistent with how labels are done. Also takes care of edit summaries.

Diff where it worked: https://www.wikidata.org/w/index.php?title=Q4115189&diff=prev&oldid=459998045

Task: https://phabricator.wikimedia.org/T159453


importer/Uploader.py
```diff
@@ -43,16 +43,14 @@ def add_labels(self, target_item, labels, log):
...
-            if (not target_item.descriptions or
-                    lang not in target_item.descriptions):
-                descs = {lang: name}
-                target_item.editDescriptions(descs)
-                pywikibot.output("Added description: " + name)
-                if log:
-                    t_id = target_item.getID()
-                    log.logit(t_id + " ADDED DESCRIPTION " + lang + " " + name)
+            self.wdstuff.add_description(
+                lang, desc_content, target_item, summary=self.summary)
```
lokal-profil: You know that you can set summary when you initialise wdstuff instead of passing it each time?

importer/Uploader.py
```diff
@@ -43,16 +43,14 @@ def add_labels(self, target_item, labels, log):
...
+            if log:
+                t_id = target_item.getID()
+                log.logit(t_id + " ADDED DESCRIPTION " +
```
lokal-profil: Use string formatting instead of concatenating the strings. `.format()` is probably the preferred but I'm also ok with `%`-formatting if that is what you use elsewhere.

importer/Uploader.py

importer/Uploader.py
```diff
@@ -46,11 +46,12 @@ def add_descriptions(self, target_item, descriptions, log):
...
+                lang, desc_content, target_item)
             if log:
                 t_id = target_item.getID()
-                log.logit(t_id + " ADDED DESCRIPTION " +
-                          lang + " " + desc_content)
+                message = "{} ADDED DESCRIPTION {} {}".format(
+                    t_id, lang, desc_content)
+                log.logit(message)
 
     def add_aliases(self, target_item, aliases, log):
```
lokal-profil: Related to this. Here you access `editAliases` directly which means that again the edit summary  does not get added. You should be able to just use `self.wdstuff.addLabelOrAlias` again (assuming add_labels, has been run before).

Or are there situations where you would on purpose want an alias without having a label?

Vesihiisi: IIRC I did it with the BBR dataset in mind, where an item can have both a name and a [fastighetsbeteckning](https://github.com/Vesihiisi/COH-tools/blob/master/importer/SeBbrSv.py#L126). I figured that I'd always want the latter as an alias, never as a label -- but relying on addLabelOrAlias seemed a bit tricky, as the dictionary passed to the Uploader could have its keys in whatever order, and thus I wasn't sure I could guarantee that the _actual_ label would be processed first.

lokal-profil: OK. The downside is that if you run them separately there is nothing preventing you from adding something to Alias which is the same as something in label. I would say it's probably best to use addLabelOrAlias as default (combining them both) and overriding it only in cases where you need to make a distinction.

For those cases you can fix the summary issue by simply adding `summary=edit_summary` to `editAliases`.

Vesihiisi: Right. In order to force-mark something as an alias, one has to use [Monument.add_alias()](https://github.com/Vesihiisi/COH-tools/blob/master/importer/Monument.py#L137). I have combined add_labels and add_aliases, which checks for those and works on them separately using editAliases, and uses an edit summary that looks the same as when using addLabelOrAlias.

importer/Uploader.py
```diff
@@ -28,17 +28,49 @@ def make_aliases(self):
...
             self.wdstuff.addLabelOrAlias(
-                lang, name, target_item)
+                language, label_content, target_item)
             if log:
                 t_id = target_item.getID()
-                message = "{} ADDED LABEL {} {}".format(t_id, lang, name)
+                message = "{} ADDED LABEL {} {}".format(
+                    t_id, language, label_content)
                 log.logit(message)
+        if len(aliases) > 0:
```
lokal-profil: If you can guarantee that aliases is an iterable then you can skip this step.

If you cannot guarantee it I would probably recommend changing the order of the variables to have `labels` and `aliases` as the last two and have both default to `None`. Then start the function with `labels = labels or {}` and same for aliases. That way you don't have to pass empty values. Alternatively leave as is but add `aliases = aliases or {}` (and optionally `labels = labels or {}`) to the start to allow aliases (and optionally labels) to be `None`.

In both cases you can skip the `len` test

importer/Uploader.py
```diff
@@ -28,17 +28,49 @@ def make_aliases(self):
...
+        However, the data model supports an optional [aliases]
+        field as well.
+        Its contents will be added directly as an alias, without checking
+        if it could be a label first.
+        """
         for label in labels:
             target_item.get()
-            name = labels[label]['value']
-            lang = labels[label]['language']
+            label_content = labels[label]['value']
```
lokal-profil: You can just use `label` here rather than going back to `labels[label]`

importer/Uploader.py
```diff
@@ -28,17 +28,49 @@ def make_aliases(self):
...
+        """
+        Add labels and aliases.
+
+        Normally, if the WD item already has a label in $lang
+        and the data object has another one, the new one
+        will be automatically added as an alias. Otherwise
+        (no existing label), it will be added as a label.
+
+        However, the data model supports an optional [aliases]
+        field as well.
```
lokal-profil: I would remove the line break here for clarity.

importer/Uploader.py
Vesihiisi: It's always an iterable -- it's [instantiated as an empty dict](https://github.com/Vesihiisi/COH-tools/blob/master/importer/Monument.py#L288), so even if there's no content in it, that empty dict is passed to this stage.

Removed the len test here and also in another place with the same logic.

importer/Uploader.py

importer/Uploader.py

PR #41, opened 2017-03-10
Create site instance once and then pass it to Uploader

Vesihiisi:
Avoid creating a new site instance for every item. Instead, create it once per batch and then pass to Uploader.

Task: https://phabricator.wikimedia.org/T159948


importer/Uploader.py
```diff
@@ -296,7 +296,7 @@ def __init__(self, monument_object, log=None, tablename=None, live=False):
         if log is not None:
             self.log = log
         self.data = monument_object.wd_item
-        site = pywikibot.Site("wikidata", "wikidata")
+        site = site
```
lokal-profil: You don't need this line (which doesn't do anything). what you want to do is set `self.repo = site`.

Right now you allow `site=None` if you do that then you need a fallback (so `self.repo = site or utils.create_site_instance("wikidata", "wikidata")`).
A better solution would be to make site an eariler parameter (just before log) and require it to have a value. I would also rename it "repo" to make it clear that it needs to be a data repository.
Maybe also add a docstring explaining the parameters.

importer/importer_utils.py
```diff
@@ -412,7 +412,7 @@ def get_random_list_sample(some_list, amount):
 
 def get_P31(q_number):
```
lokal-profil: I would make this take repo/site as a parameter.

importer/importer.py
```diff
@@ -196,6 +196,7 @@ def get_items(connection,
         print("USING RANDOM SAMPLE OF " + str(short))
     filename = specific_table_name + "_" + utils.get_current_timestamp()
     problem_reports = []
+    wikidata_site = utils.create_site_instance("wikidata", "wikidata")
     for row in database_rows:
         monument = class_to_use(row, mapping, data_files, existing)
```
lokal-profil: Add repo as an argument in Monument.__init__, and pass it here. That way you can use it for calls to e.g. `get_P31`.

importer/importer.py
```diff
@@ -196,6 +196,7 @@ def get_items(connection,
         print("USING RANDOM SAMPLE OF " + str(short))
     filename = specific_table_name + "_" + utils.get_current_timestamp()
     problem_reports = []
+    wikidata_site = utils.create_site_instance("wikidata", "wikidata")
```
lokal-profil: You might consider also creating the commons site which you also use in Uploader.

The Wikipedia sites are probably not worth it since they differ too much.

importer/importer_utils.py
```diff
@@ -439,3 +439,13 @@ def create_wlm_url(country, language, id):
     url_base = "https://tools.wmflabs.org/heritage/api/api.php?action=search&format=json&srcountry={}&srlanguage={}&srid={}"
     return url_base.format(
         country, language, quote(id))
+
+
+def create_site_instance(language, family):
+    """
+    Create an instance of a Wiki site.
```
lokal-profil: I would shorten this down to "Convenience function to create an instance of a Wiki site." or possibly "Create an instance of a Wiki site (convenience function)."

importer/Uploader.py
Vesihiisi: Done (as earlier parameter).

importer/Uploader.py
```diff
@@ -282,7 +282,22 @@ def set_wd_item(self):
...
+    def __init__(self,
+                 monument_object,
+                 repo,
+                 log=None,
+                 tablename=None,
+                 live=False):
+        """
+        Initialize an Upload object for a single Monument.
+
+        @param: monument_object Dictionary of Monument data
```
lokal-profil: the format `@param param_name: explanation` is more common

importer/Uploader.py
lokal-profil: This bit looks good.

lokal-profil:
Note that I just looked at the first bit since I don't think you've done the other bits (if you opted not to do them then leave a comment in each)

importer/Uploader.py
lokal-profil: or use `:param param_name: explanation`

Vesihiisi:
It now creates a repo in importer and passes it to both Monument (to interact with importer_utils) and Uploader.

importer/Monument.py
```diff
@@ -295,7 +295,7 @@ def print_datafiles_titles(self):
         for title in self.data_files:
             print(title)
 
-    def __init__(self, db_row_dict, mapping, data_files, existing):
+    def __init__(self, db_row_dict, mapping, data_files, existing, repository):
```
lokal-profil: Does this have a docstring and if so does that mention the repository parameter? (Hard to say with all of the parallel patches)

tests/test_importer_utils.py
```diff
@@ -1,6 +1,7 @@
+import importer.importer_utils as utils
```
lokal-profil: local imports should go last

importer/importer_utils.py
```diff
@@ -439,3 +438,8 @@ def create_wlm_url(country, language, id):
     url_base = "https://tools.wmflabs.org/heritage/api/api.php?action=search&format=json&srcountry={}&srlanguage={}&srid={}"
     return url_base.format(
         country, language, quote(id))
+
+
+def create_site_instance(language, family):
```
lokal-profil: So there is a sneaky thing we can do here which might make life a bit easier. We make this function cache any created sites.

If we at the top of utils introduces `site_cache = {}`
then here we can use.
```python
site_key = (language, family)
site = site_cache.get(site_key)
if not site:
    site = pywikibot.Site(language, family)
    site_cache[site_key] = site
return site
```

tests/test_importer_utils.py
Vesihiisi: DOne.

importer/Monument.py
Vesihiisi: It got a docstring in #49 , but I added an updated one now with the extra parameter.

importer/importer_utils.py

importer/importer_utils.py
```diff
@@ -137,22 +139,29 @@ def count_wikilinks(text):
...
         try:
             item = pywikibot.ItemPage.fromPage(page)
-            return item.getID()
         except pywikibot.NoPage:
-            print("Failed to get page for {} - {}."
-                  "It probably does not exist.".format(
-                      language, page_title)
-                  )
-            return
+            if page.namespace() == 0:  # main namespace
```
lokal-profil: While touching this the following might be simpler:
```python
if page.namespace() != 0:  # main namespace
    return
item = wdstuff.make_new_item_from_page(page, summary)
```

lokal-profil: No need for the `else`

importer/SeBbrSv.py
```diff
@@ -165,8 +165,8 @@ def set_architect(self):
             for name in architects:
                 wp_page = name.title
                 q_item = utils.q_from_wikipedia("sv", wp_page)
-                if q_item:
-                    if utils.is_whitelisted_P31(q_item, ["Q5"]):
+                if q_item is not None:
```
lokal-profil: I think this was reverted by mistake.

importer/importer_utils.py
```diff
@@ -147,22 +148,21 @@ def q_from_wikipedia(language, page_title):
...
     """
-    site = pywikibot.Site(language, "wikipedia")
-    repo = site.data_repository()
-    page = pywikibot.Page(site, page_title)
+    wp_site = pywikibot.Site(language, "wikipedia")
+    page = pywikibot.Page(wp_site, page_title)
     summary = "Creating item for {} on {}wp."
     summary = summary.format(page_title, language)
-    wdstuff = wds(repo, edit_summary=summary)
+    wdstuff = wds(create_site_instance("wikidata", "wikidata"), edit_summary=summary)
```
lokal-profil: I would split this into two statements. `repo = ... `and `wdstuff = wds(repo, ...`

importer/importer_utils.py
```diff
@@ -147,22 +148,21 @@ def q_from_wikipedia(language, page_title):
...
             page = page.getRedirectTarget()
         try:
             item = pywikibot.ItemPage.fromPage(page)
         except pywikibot.NoPage:
-            if page.namespace() == 0:  # main namespace
-                item = wdstuff.make_new_item_from_page(page, summary)
-            else:
+            if page.namespace() != 0:  # main namespace
                 return
+            else:
```
lokal-profil: you don't need this `else` (because of the `return` above)

importer/importer_utils.py

importer/SeBbrSv.py

PR #42, opened 2017-03-13
Final preview page of se-ship

Vesihiisi:
Finalize the data processing in order to present a preview page.

https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/se-ship_(sv)/preview

Includes:

- Sourcing the ship type field
- Adding descriptions

Task: https://phabricator.wikimedia.org/T160326


importer/SeShipSv.py
```diff
@@ -118,6 +119,13 @@ def set_monuments_all_id(self):
         """
         self.monuments_all_id = self.signal
 
+    def set_descriptions(self):
+        """Set default descriptions."""
+        DESC_BASES = {"sv": "K-märkt fartyg",
```
lokal-profil: I believe we can clarify this further by saying "Kulturmärkt bruksfartyg" [1] to distinguish it even more from "Kulturmärkta fritidsbåtar" [2]. (Both of which are currently mentioned on [3] which is the wp page for the protection status)

[1] https://sv.wikipedia.org/wiki/Lista_%C3%B6ver_kulturm%C3%A4rkta_bruksfartyg_i_Sverige
[2] https://sv.wikipedia.org/wiki/Lista_%C3%B6ver_k-m%C3%A4rkta_fritidsb%C3%A5tar
[3] https://sv.wikipedia.org/wiki/Kulturm%C3%A4rkning_av_fartyg_i_Sverige

lokal-profil: If you are not using this as a global then I would use a lower case variable name

Vesihiisi: Done (I guess this assumes that all the items in the db are of this type, but it doesn't look like any of the items in [2] are also included in [1]).

Vesihiisi: Done.

PR #46, opened 2017-03-16
[se-fornmin] Finalize processing code

Vesihiisi:
Make sure that the preview page is ready for release:

https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/se-fornmin_(sv)/preview

Add reporting of unmatched values.

Task: https://phabricator.wikimedia.org/T160617


importer/SeFornminSv.py
```diff
@@ -138,17 +139,6 @@ def set_location(self):
             self.add_statement("location", self.get_socken(
```
lokal-profil: Should there be an `add_to_report` here in case matching fails?

Vesihiisi: Added it.

Vesihiisi: Re: 1 -- I'm not sure which part exactly you mean by 'parser'?

lokal-profil: I mean in `utils.get_wikilinks()` which in turn calls `mediawikiparserfromhell`. So if you send it a string without `[[` it should just return an empty list?

Vesihiisi: Ah, right. I've cleaned it up.

PR #47, opened 2017-03-16
[se-bbr] Finalize processing code

Vesihiisi:
Make sure that the preview page is ready for release:

https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/se-bbr_(sv)/preview

Add reporting of unmatched values.

Task: https://phabricator.wikimedia.org/T160625


importer/SeBbrSv.py
```diff
@@ -50,7 +50,8 @@ def set_heritage_bbr(self):
         """
         protection_date = False
         url = "http://kulturarvsdata.se/" + \
-            self.wd_item["statements"][self.props["cultural_heritage_sweden"]][0]["value"]
+            self.wd_item["statements"][
```
lokal-profil: I would break this out as a separate varaible `kulturarvsdata_id`

importer/SeBbrSv.py
```diff
@@ -81,15 +82,6 @@ def set_heritage_bbr(self):
             qualifier = None
         self.add_statement("heritage_status", type_q, qualifier)
```
lokal-profil: the reference for this should be kulturarvsdata ideally something like:
* dataset= BBR/kulturarvsdata
* id = kulturarvsdata_id
* url = url
* retrieval date = today

importer/SeBbrSv.py
lokal-profil: _I cannot comment on the bits below inline so here goes:_

* I would also add an example of what ns5:spec might look like to explain the string splitting logic.
* On line 70 below. Use a normal comment rather than a multiline string.
* What happens if `utils.date_to_dict` cannot interpret the string as a date?, a try/except would be nice here.

importer/SeBbrSv.py
```diff
@@ -119,7 +114,10 @@ def set_location(self):
         if self.has_non_empty_attribute("plats"):
             if utils.count_wikilinks(self.plats) == 1:
                 location = utils.q_from_first_wikilink("sv", self.plats)
```
lokal-profil: Do we want some filtering here on the type of object?

importer/SeBbrSv.py
```diff
@@ -183,7 +184,12 @@ def set_inception(self):
         if self.has_non_empty_attribute("byggar"):
             year_parsed = utils.parse_year(self.byggar)
             if year_parsed is not None:
```
lokal-profil: no need for "is not None". Just use `if year_parsed` (since "0" is not a valid value anyway it shouldn't cause an issue)

importer/SeBbrSv.py
```diff
@@ -141,9 +139,11 @@ def set_no_of_buildings(self):
         extracted_no = utils.get_number_from_string(
             utils.get_text_inside_brackets(self.funktion))
         if extracted_no is not None:
```
lokal-profil: no need for "is not None". Just use `if extracted_no` (since "0" should also be skipped)

importer/SeBbrSv.py
```diff
@@ -104,7 +96,10 @@ def set_architect(self):
                 q_item = utils.q_from_wikipedia("sv", wp_page)
                 if q_item is not None:
```
lokal-profil: just use `if q_item:`

importer/SeBbrSv.py
```diff
@@ -81,15 +82,6 @@ def set_heritage_bbr(self):
             qualifier = None
         self.add_statement("heritage_status", type_q, qualifier)
 
-    def set_function(self):
```
lokal-profil: How come this was removed?

importer/SeBbrSv.py
```diff
@@ -141,9 +139,11 @@ def set_no_of_buildings(self):
         extracted_no = utils.get_number_from_string(
             utils.get_text_inside_brackets(self.funktion))
         if extracted_no is not None:
+            ref = self.wlm_source
             self.add_statement(
                 "has_parts_of_class", "Q41176",
-                {"quantity": {"quantity_value": extracted_no}})
+                {"quantity": {"quantity_value": extracted_no}},
+                refs=[ref])
```
lokal-profil: Two questions on `add_statement`:
1) Do you ever need more than one reference? (e.g. WikidataStuff does not support it)
2) If you do need multiple are single references more frequent? If so you might consider adding a `listify` call in `add_statements`. I.e. something like:
```
if refs and not isinstance(refs, list):
    refs = [refs, ]
```
or reuse the function for this in WikidataStuff (`helpers.listify()`)

lokal-profil: And taking this one step further. Would it be possible to use `self.wlm_source` as a default?
I'm thinking something like:
```
def add_statement(..., refs=None, ...):
    if refs is None or refs is True:
        refs = self.wlm_source
    elif refs is False:
        refs = None
    ...
```
That way you can still opt not to use a ref (e.g. for images) using `refs=False` but by default it would use `refs=self.wlm_source`. Or you can make `refs=True` trigger the default behaviour and require that `refs` always takes a value (It would then allow True/False, a ref or a list of refs).

lokal-profil: Note also that using `{}` and `[]` as default arguement values (as is done in `add_statements` today) is discouraged (see https://www.quantifiedcode.com/app/issue_class/3P0qV6OB)

lokal-profil: OK. I'd break this out as a separate task and not have it block this task but have it as a blocker for running the fmis and bbr uploads.

Vesihiisi: I broke this out as https://phabricator.wikimedia.org/T161303

Vesihiisi: Re: 1 -- I have only used single references so far, but I erred on the side of caution when making this... It does make sense to optimize for single references though.

importer/SeBbrSv.py
Vesihiisi: - variable: Done.
- example: Done.
- comment: Done.

importer/SeBbrSv.py

importer/SeBbrSv.py

importer/SeBbrSv.py

Vesihiisi:
I have moved back the 'function' parser, and enabled filtering of location by type.

I have added a reference for the heritage status with a link to kulturarvsdata, it can be seen [here](https://www.wikidata.org/w/index.php?title=Q4115189&oldid=470860434).

As for the add_statement, it's been moved to PR #50 -- I guess it should be merged first, so that we can remove the superfluous references here (since #50 enables default references).

importer/SeBbrSv.py
```diff
@@ -49,38 +49,87 @@ def set_heritage_bbr(self):
...
+            except ValueError:
+                self.add_to_report("protection_type", bbr_type)
         else:
             qualifier = None
-        self.add_statement("heritage_status", type_q, qualifier)
+        kulturarv_source = self.create_kulturarv_source(url)
+        self.add_statement(
+            "heritage_status", type_q, qualifier, refs=[kulturarv_source])
+
+    def create_kulturarv_source(self, url):
```
lokal-profil: Needs docstring

importer/SeBbrSv.py
```diff
@@ -49,38 +49,87 @@ def set_heritage_bbr(self):
...
             # 1969-01-31
-            date_dict = utils.date_to_dict(protection_date, "%Y-%m-%d")
-            qualifier = {"start_time":
-                         {"time_value": date_dict}}
+            try:
+                date_dict = utils.date_to_dict(protection_date, "%Y-%m-%d")
+                qualifier = {"start_time":
+                             {"time_value": date_dict}}
+            except ValueError:
+                self.add_to_report("protection_type", bbr_type)
```
lokal-profil: At this point you would still reach `self.add_statement`. Is that desired? If so you probably want to set `qualifier = None` already after line 88. If not you probably want to `return` here.

lokal-profil: Actually adding `qualifier = None` after line 88 (and skipping the else on line 97) might be clearer even if you want to `return` here.

importer/SeBbrSv.py
```diff
@@ -49,38 +49,87 @@ def set_heritage_bbr(self):
...
+            "heritage_status", type_q, qualifier, refs=[kulturarv_source])
+
+    def create_kulturarv_source(self, url):
+        source_item = self.sources["bebyggelseregistret"]
+        today = utils.today_dict()
+        prop_reference_url = self.props["reference_url"]
+        prop_date = self.props["retrieved"]
+        prop_stated = self.props["stated_in"]
+        prop_reference_url = self.props["reference_url"]
+        kulturarv_source = {"source":
```
lokal-profil: I would linebreak this as follows to make it more legible:
```
kulturarv_source = {
    "source": {"prop": prop_stated, "value": source_item},
    "published": {"prop": prop_date, "value": today},
    "reference_url": {"prop": prop_reference_url, "value": url}}
```

importer/SeBbrSv.py
```diff
@@ -49,38 +49,87 @@ def set_heritage_bbr(self):
...
+                            {"prop": prop_stated, "value": source_item},
+                            "published": {"prop": prop_date, "value": today},
+                            "reference_url": {"prop": prop_reference_url,
+                                              "value": url}}
+        return kulturarv_source
+
+    def set_function(self):
```
lokal-profil: Needs docstring

importer/SeBbrSv.py
```diff
@@ -49,38 +49,87 @@ def set_heritage_bbr(self):
...
+        functions_map = self.data_files["functions"]["mappings"]
+        if self.has_non_empty_attribute("funktion"):
+            functions = utils.get_rid_of_brackets(
```
lokal-profil: I would linebreak this one as follows to make it clearer
```
functions_string = utils.get_rid_of_brackets(self.funktion)
functions = functions_string.lower().split(",")
```

importer/SeBbrSv.py
```diff
@@ -49,38 +49,87 @@ def set_heritage_bbr(self):
...
+                self.funktion).lower().split(",")
+            for item in functions:
+                function = item.strip()
+                try:
+                    function_item = [functions_map[x]["items"]
```
lokal-profil: I would separate the list comprehension from the get [0][0] of that list.

importer/SeBbrSv.py
```diff
@@ -94,28 +143,43 @@ def set_architect(self):
...
         """
+        settlements_dict = self.data_files["settlements"]
         if self.has_non_empty_attribute("plats"):
             if utils.count_wikilinks(self.plats) == 1:
-                location = utils.q_from_first_wikilink("sv", self.plats)
-                ref = self.wlm_source
-                self.add_statement("location", location, refs=[ref])
+                location_q = utils.q_from_first_wikilink("sv", self.plats)
+                try:
+                    legit_location = [x["item"] for x in settlements_dict if x["item"] == location_q][0]
```
lokal-profil: I would separate the list comprehension from the get [0] of that list.

importer/SeBbrSv.py
```diff
@@ -165,11 +227,10 @@ def set_adm_location(self):
         try:
             municipality = [x["item"] for x in municip_dict if x[
                 "en"].lower() == pattern_en][0]
-            ref = self.wlm_source
-            self.add_statement("located_adm", municipality, refs=[ref])
+            self.add_statement("located_adm", municipality)
         except IndexError:
             print("Could not parse municipality: {}.".format(self.kommun))
-            self.add_to_report("kommun", self.kommun)
+            return
```
lokal-profil: Why do we no longer add this to the report?

Vesihiisi:
Note: has to be updated to use the latest modifications in add_statement and item matching.

importer/SeBbrSv.py

importer/SeBbrSv.py

importer/SeBbrSv.py

importer/SeBbrSv.py

importer/SeBbrSv.py
Vesihiisi: I guess this depends on whether we see a heritage status with a malformed date as valuable. I think if the status looks OK but the date is skipped it can still be added.

I've moved the `qualifier = None` up.

importer/SeBbrSv.py

importer/SeBbrSv.py

importer/SeBbrSv.py
Vesihiisi: Done (it got reverted to a previous version for some reason).

Vesihiisi:
Both add_statement and item matching are compatible with master.

As for the ugly docstrings, these are getting fixed in #54.

importer/Uploader.py
```diff
@@ -204,12 +204,15 @@ def add_claims(self, wd_item, claims, log):
         if wd_item:
             for claim in claims:
                 prop = claim
+                print(prop)
```
lokal-profil: Don't commit with test/debug code.

importer/Uploader.py
```diff
@@ -220,16 +223,17 @@ def add_claims(self, wd_item, claims, log):
...
-                            If we have references of different sort,
-                            this will have to be appended.
-                            """
-                            if utils.is_valid_url(ref):
-                                ref = self.make_url_reference(ref)
-                            else:
-                                ref = self.make_stated_in_reference(ref)
+                        if refs:
+                            for ref in refs:
+                                """
```
lokal-profil: This should be a normal comment, not a multiline string.

lokal-profil: But might be fixed in another patch.

importer/Uploader.py
```diff
@@ -220,16 +223,17 @@ def add_claims(self, wd_item, claims, log):
...
-                            """
-                            This only works if it's a url.
-                            If we have references of different sort,
-                            this will have to be appended.
-                            """
-                            if utils.is_valid_url(ref):
-                                ref = self.make_url_reference(ref)
-                            else:
-                                ref = self.make_stated_in_reference(ref)
+                        if refs:
```
lokal-profil: The alternative would be to set `refs = []` on [L119](https://github.com/Vesihiisi/COH-tools/commit/669e78b0fb0044e1abc4656aefccc74fb59abde2#diff-3c84fc200ee0223310cef2e0937815beR119). That wouldn't require any other changes and might actually be clearer (since `refs=None` before that line is what triggers default ref)

importer/Uploader.py
Vesihiisi: Fixed in #54

importer/Uploader.py

importer/Uploader.py

PR #49, opened 2017-03-24
Existing check

Vesihiisi:
Connect the data object to possible existing WD item.

Use:
1. An optional unique property given in the mapping file
2. The "wd_item" column in the db
3. The "monument_article" column in the db.

Task: https://phabricator.wikimedia.org/T161114


importer/Monument.py
```diff
@@ -262,49 +258,70 @@ def create_wlm_source(self, monuments_all_id):
...
+        if wd_item_keyword in self.raw_data:
+            return self.raw_data[wd_item_keyword]
+        else:
+            return None
+
     def exists_with_prop(self, mapping):
-        if self.existing is None:
-            return
+        """
+        Check whether there already exists an item
```
lokal-profil: First sentence of docstring needs to fit on a single line.

importer/Monument.py
```diff
@@ -262,49 +258,70 @@ def create_wlm_source(self, monuments_all_id):
...
-                    pass  # match
-                self.set_wd_item(wd_item)
+                return wd_item
             else:
-                print("There's no item with {} = {} on Wikidata.".format(
-                    unique_prop, val_to_check))
+                return None
+
+    def in_known_items(self, wd_item):
+        """Check whether given item is on the list of items with certain prop"""
```
lokal-profil: I would say _in_ the list

importer/Monument.py
```diff
@@ -262,49 +258,70 @@ def create_wlm_source(self, monuments_all_id):
...
+        return wd_item in self.existing.items()
+
+    def find_matching_wikidata(self, mapping):
```
lokal-profil: Needs docstring

importer/Monument.py
```diff
@@ -262,49 +258,70 @@ def create_wlm_source(self, monuments_all_id):
...
+        item = self.exists_with_prop(mapping)
+        if not item:
+            item = self.exists_with_wd_item()
+            if not item:
+                item = self.exists_with_monument_article(self.mapping["language"])
+            if item and self.in_known_items(item):
+                self.upload = False
```
lokal-profil: Does this also trigger some type of report?

importer/Monument.py
lokal-profil: Is there a need to have this return the `wd_item` rather than always passing it on to `self.set_wd_item()`?

importer/Monument.py
```diff
@@ -262,49 +258,70 @@ def create_wlm_source(self, monuments_all_id):
...
         self.wd_item["descriptions"] = {}
         self.wd_item["wd-item"] = None
-        self.mapping = mapping.file_content
 
     def print_datafiles_titles(self):
         """Print out the keys in self.data_files."""
         for title in self.data_files:
             print(title)
 
     def __init__(self, db_row_dict, mapping, data_files, existing):
```
lokal-profil: A docstring here defining the `existing` parameter (well all of them) would be good since we interact with it in `in_known_items()`

importer/SeArbetslSv.py
```diff
@@ -135,28 +135,29 @@ def set_is(self):
         ref = self.arbetsam_source
         self.add_statement("is", arbetslivsmuseum, refs=[ref])
 
+    def exists_with_monument_article(self, language):
```
lokal-profil: Add a short docstring explaining you are overriding with "sv, monuments_article"

importer/Monument.py
```diff
@@ -262,49 +258,70 @@ def create_wlm_source(self, monuments_all_id):
...
-                self.set_wd_item(wd_item)
+                return wd_item
             else:
-                print("There's no item with {} = {} on Wikidata.".format(
-                    unique_prop, val_to_check))
+                return None
+
+    def in_known_items(self, wd_item):
+        """Check whether given item is on the list of items with certain prop"""
+        return wd_item in self.existing.items()
```
lokal-profil: is `wd_item` the key used in `self.existing`? If so `wd_item in self.existing` should be enough.

importer/Monument.py
```diff
@@ -66,9 +66,11 @@ def print_wd_to_table(self):
                 elif "quantity_value" in value:
                     value_to_print += str(value["quantity_value"])
                     if "unit" in value:
-                        value_to_print += " " + utils.wd_template("Q", value["unit"])
+                        value_to_print += " " + \
```
Vesihiisi: Done.

importer/Monument.py
```diff
@@ -262,49 +258,70 @@ def create_wlm_source(self, monuments_all_id):
                 "reference_url": {"prop": prop_reference_url,
                                   "value": self.wlm_url}}
 
+    def exists_with_monument_article(self,
+                                     language,
+                                     article_keyword="monument_article"):
+        """Check whether the data object has anythin in the monument_article field."""
```
Vesihiisi: Done (in all the instances),.

importer/Monument.py
Vesihiisi: Done.

importer/Monument.py
Vesihiisi: Done.

importer/SeArbetslSv.py
Vesihiisi: Done.

importer/Monument.py
Vesihiisi: Added a report.

importer/Monument.py
Vesihiisi: Done.

importer/Monument.py
Vesihiisi: No, it's the other way round -- e.g. `{'1855': 'Q28934220', …}`, i.e. the value is the key and the Q-item is the item.

importer/Monument.py
Vesihiisi: Added docstring.

Vesihiisi: I think this makes it easier to see how it's matched, given it's a fairly complicated process.

importer/Monument.py
```diff
@@ -261,7 +264,7 @@ def create_wlm_source(self, monuments_all_id):
     def exists_with_monument_article(self,
                                      language,
                                      article_keyword="monument_article"):
-        """Check whether the data object has anythin in the monument_article field."""
+        """Get the Q-item connected to monument_article (or equivalent), if any."""
```
lokal-profil: Use either "Wikidata item" or "Qid/Q-id of item" since Q-item isn't a term used elsewhere.

importer/Monument.py
lokal-profil: ok

importer/Monument.py
lokal-profil: ok

importer/SeArbetslSv.py
```diff
@@ -136,6 +136,7 @@ def set_is(self):
         self.add_statement("is", arbetslivsmuseum, refs=[ref])
 
     def exists_with_monument_article(self, language):
+        """Override default behavior with unusual field name."""
```
lokal-profil: Don't we actually just set the language? Since "monument_article" is the default?

Vesihiisi: Do you mean that the second parameter should be left out?

importer/Monument.py
Vesihiisi: Done.

importer/SeArbetslSv.py
lokal-profil: I meant for the docstring (although yes I guess we could also leave out the second param).

Vesihiisi: Done.

PR #52, opened 2017-03-29
Add linting and TravisCI capabilities

lokal-profil:
Adds flake8 and pydocstyle linting along with unit test running using tox.

Also adds the needed connections to make TravisCI run this using tox.

Task: https://phabricator.wikimedia.org/T159116


lokal-profil:
The failing tests are due to:
# P31 changing in Q146 (needs fixing in `tests/test_importer_utils.py:test_get_P31_two`)
# Various docstring errors throughout the country files.

I would suggest that both of these be addressed by a separate patch.

PR #53, opened 2017-03-30
Create a new WD item when encountering page without one

Vesihiisi:
When the data item is connected to a WP article without a Wikidata item, create one automatically.

Task: https://phabricator.wikimedia.org/T161711


lokal-profil:
Wds must be instantiated (with a repo object) to be used.

Vesihiisi:
Done.

lokal-profil:
Would be nice to pass it a repo/wds object but I'm ok with this.

lokal-profil:
Would be nice to pass it a repo/wds object but that is #41 so I'm ok with this for now.

PR #57, opened 2017-03-31
Improve socken matching logic

Vesihiisi:
Improve the socken matching so that even socken with unusual names (i.e. without the word socken in the name) get matched to correct WD items (that have P777).

Task: https://phabricator.wikimedia.org/T161886


importer/SeFornminSv.py
```diff
@@ -145,12 +145,18 @@ def set_location(self):
...
             socken = self.get_socken(self.socken, self.landskap)
-            if socken is not None:
+            if socken:
                 self.add_statement("location", socken)
             else:
-                raw_socken = "{} ({})".format(self.socken, self.landskap)
-                self.add_to_report("socken", raw_socken)
+                try:
+                    socken = [x["item"] for x in socken_dict
+                              if x["itemLabel"].startswith(self.socken)][0]
```
lokal-profil: if you are going via Label I would do:
```python
socken_candidate = [x["item"] for x in socken_dict
                                  if x["itemLabel"].startswith(self.socken)]
if len(socken_candidate != 1):
    raise ValueError
self.add_statement("location", socken_candidate[0])
```
Because I don't trust that there isn't at least one case of two of them starting with the same name.

importer/importer.py
```diff
@@ -95,6 +95,7 @@ def create_connection(arguments):
                    {"class": SeFornminSv,
                     "data_files":
                     {"municipalities": "sweden_municipalities.json",
+                    "socken": "sweden_socken.json",
```
lokal-profil: indent one space

importer/SeFornminSv.py
Vesihiisi: Done.

PR #58, opened 2017-03-31
Get text inside outermost brackets

Vesihiisi:
Current behavior of get_text_inside_brackets():

`(foo (bar (cat)))` → `foo (bar (cat`

expected behavior:

`(foo (bar (cat)))` → `foo (bar (cat))`

I.e. only get what's inside the outermost set of brackets.

Task: https://phabricator.wikimedia.org/T161894


tests/test_importer_utils.py
```diff
@@ -79,6 +79,16 @@ def test_get_text_inside_brackets_succeed(self):
...
 
+    def test_get_text_inside_brackets_nested_one(self):
+        text = "outsidetext (foo (cat))"
+        output = "foo (cat)"
+        self.assertEqual(utils.get_text_inside_brackets(text), output)
+
+    def test_get_text_inside_brackets_nested_two(self):
+        text = "text (foo (bar (cat))) around"
+        output = "foo (bar (cat))"
+        self.assertEqual(utils.get_text_inside_brackets(text), output)
```
lokal-profil: Add a test for `pre (foo) mid (bar) post`.
Expected behaviour (in my opinion) is one of the following:

1. Return only `foo`
2. Return `(foo, bar)` (or `[foo, bar]`)
3. Raise an error

What I believe will happen right now is `foo) mid (bar`

Vesihiisi: It does return `"foo) mid (bar"`.

I guess in case of multiple instances it makes most sense to return a list, so I'll try going with that.

Vesihiisi: Even though it can theoretically lead to cases with both multiple instances and nested brackets, like:

`foo (one) bar (two (and a half)) cat (three)` → `[one, two (and a half), three]`

Vesihiisi: ok this wasn't a good idea in any way.

Vesihiisi: So i made it error out instead.

importer/importer_utils.py
```diff
@@ -306,7 +306,15 @@ def get_text_inside_brackets(text):
     if "(" in text:
         first_bracket_index = text.find('(')
         last_bracket_index = text.rfind(')')
-        return text[first_bracket_index + 1:last_bracket_index]
+        result = text[first_bracket_index + 1:last_bracket_index]
+        if "(" in result or ")" in result:
+            for character in result:
```
lokal-profil: Why is "(" ok when ")" isn't?

I would probably raise ValueError on either so:
```python:
if "(" in result or ")" in result:
    raise ValueError('Nested brackets encountered')
return result
```

You should also update the docstring to clarify that the function can only handle one or less sets of brackets.

tests/test_importer_utils.py
```diff
@@ -91,8 +91,7 @@ def test_get_text_inside_brackets_nested_two(self):
 
     def test_get_text_inside_brackets_twice(self):
         text = "pre (foo) mid (bar) post"
-        output = "foo"
-        self.assertEqual(utils.get_text_inside_brackets(text), output)
+        self.assertRaises(ValueError, utils.get_text_inside_brackets, text)
```
lokal-profil: Could you also add a test for `"pre (foo"` I believe it will not give the expected result (either `ValueError` or "foo").

This comes about because find returns `-1` if not found.
You can get around this (and similar problems with strings such as `"foo )bar( cat"` and `"foo) bar (cat)"`) by:
```python
first_bracket_index = text.find('(')
last_bracket_index = text[first_bracket_index:].rfind(')')
if last_bracket_index < 0 or text[:first_bracket_index].rfind(')') > 0:
     raise ValueError('Unmatched brackets encountered')
result = text[first_bracket_index + 1:first_bracket_index+last_bracket_index]
```

importer/importer_utils.py
Vesihiisi: "(" is okay because it's okay to have some brackets in the result, if we allow for nested brackets. But if the first bracket in the result is a closing one -- `foo)` -- then something went wrong.

Vesihiisi: Assuming we do allow for nested brackets.

importer/importer_utils.py
```diff
@@ -302,19 +303,30 @@ def get_rid_of_brackets(text):
...
-                if character == "(":
-                    return result
-                elif character == ")":
-                    raise ValueError()
-        else:
-            return result
+        last_bracket_index = text[first_bracket_index:].rfind(')')
+        if last_bracket_index < 0 or text[:first_bracket_index].rfind(')') > 0:
+            raise ValueError("Unmatched brackets encountered.")
+        result = text[first_bracket_index +
```
lokal-profil: I would linebreak this differently or possibly add `last_bracket_index += first_bracket_index` just before if that isn't possible.

importer/importer_utils.py
```diff
@@ -302,19 +303,30 @@ def get_rid_of_brackets(text):
...
+                      1:first_bracket_index + last_bracket_index]
+        for character in result:
```
lokal-profil: If you only want to ensure that the first encountered is a "(" then I would use find again.
```python
if result.find('(') < 0 or (result.find(')') > 0 and result.find('(') > result.find(')')):
    raise ValueError("Unmatched brackets encountered.")
```
should do that although you might want to make it more legible.

importer/importer_utils.py
```diff
@@ -302,19 +303,30 @@ def get_rid_of_brackets(text):
...
-    """Get the content of the first encountered occurence of round brackets."""
+    """
+    Get the content of the first encountered occurence of round brackets.
+
+    Handles nested brackets by getting the content of
+    the first level:
+        foo (cat) → cat
+        text (foo (bar (cat))) around → foo (bar (cat))
+
+    Does not handle multiple instances of brackets on the same level.
```
lokal-profil: Add something about it raising an error on mismatching brackets as well.

Vesihiisi: done.

importer/importer_utils.py

importer/importer_utils.py
Vesihiisi: The first clause, `if result.find('(') < 0 `, would make a text-only result (`foo`) throw an error? Since it does not contain any brackets, it would give `-1`. So I've removed it.

lokal-profil: Ah. I meant for it to be triggered after "`if "(" in result or ")" in result` which I now see was removed.
The current logic would fail to complain if only a `)` is present (since `find('(')` would give `-1`).
I guess the test would be on "foo (bar))".
to fix I guess you would have to change to
`if result.find(')') > 0 and (result.find('(') > result.find(')') or result.find('(') < 0):`

Vesihiisi: Done, and added test.

importer/importer_utils.py
```diff
@@ -324,8 +324,11 @@ def get_text_inside_brackets(text):
             raise ValueError("Unmatched brackets encountered.")
         last_bracket_index += first_bracket_index
         result = text[first_bracket_index + 1:last_bracket_index]
-        if result.find(')') > 0 and result.find('(') > result.find(')'):
-            raise ValueError("Unmatched brackets encountered.")
+        if "(" in result or ")" in result:
```
lokal-profil: You can drop this conditional (since the next was augmented it already covers this as well (essentially the `result.find(')') > 0` and `result.find('(') > 0`).

Vesihiisi: ah, right. looks like that will work.

PR #59, opened 2017-04-03
Connect to database on Tool Labs

Vesihiisi:
Connect to the database on the Toollabs server.

Detect automatically if run on Toollabs or locally, and select the right credentials.

Task: https://phabricator.wikimedia.org/T160351


lokal-profil:
Review as comment die to being on phone:

Looks good just minor things:
* add docstrings to updated methods which didn't have them
* I would probably use partition instead of split for credentials.
* any reason for not set credential values in the if blocks?
* do you need the "if not on labs" test to set the arguments if they are overwritten later anyway?

Vesihiisi:
1. Done.
2. Done.
3. I find the current way easier to understand, but I've changed it.
4. I think it makes it easier to understand by making it obvious the user can't pass any input for these args if we're on labs.

lokal-profil:
On 4. Do you get an error if you pass one of those args when on labs or is it silently ignored?

Vesihiisi:
Re 4: It makes an `error: unrecognized arguments`.

PR #60, opened 2017-04-05
[se-bbr] Handle Error when parsing brackets

Vesihiisi:
Since [importer_utils.get_text_inside_brackets()](https://github.com/Vesihiisi/COH-tools/blob/master/importer/importer_utils.py#L305) can throw an exception when parsing malformed labels, make sure it doesn't kill everything.

Task: https://phabricator.wikimedia.org/T162248


lokal-profil:
So we add the label even if it is malformed? Not saying it is wrong bur it is different from other times when we add things to the report

Vesihiisi:
Yeah, I was thinking about this for a while... I think while other things can be left out, a label of some sort is a must-have, as long as we have some source data for it. In this particular case, the reason for the error will be weird parens, which is not too bad -- the label will still be readable and meaningful.

lokal-profil:
Ok.
Maybe add a small comment about that motivation/difference.

Vesihiisi:
Ok! :thumbsup:

PR #61, opened 2017-04-07
Make sure name + description is unique combination

Vesihiisi:
Create unique name + description combinations to avoid the API error when uploading.

Task: https://phabricator.wikimedia.org/T162453


importer/SeFornminSv.py
```diff
@@ -30,24 +30,23 @@ def set_descriptions(self):
...
+        which ensures its uniqueness.
         """
         DESC_BASE = "fornminne"
-        description = ""
+        landskap = self.landskap
         if len(self.typ) > 0:
-            description = self.typ.lower()
+            instance_of = self.typ.lower()
+        else:
+            instance_of = DESC_BASE
```
lokal-profil: add a linebreak after this to more clearly separate the logical blocks

Vesihiisi: Done.

PR #62, opened 2017-04-11
Add offset as command line argument

Vesihiisi:
Add `--offset <int>` as a command line argument in order to read the database table with a specific offset.

Task: https://phabricator.wikimedia.org/T162659


importer/importer.py
```diff
@@ -43,8 +43,17 @@ def __init__(self, countryname, languagename):
...
+def make_query(specific_table, offset):
+    """
+    Generate a query to retrieve data from database.
+
+    :param specific_table: Name of table to retrieve data from.
+    :param offset: Optional offset to start from.
+    """
+    query = 'select DISTINCT *  from `{}`'.format(specific_table)
+    if isinstance(offset, int):
+        query = query + " LIMIT {}, 18446744073709551615;".format(str(offset))
```
lokal-profil: `query = query + " LIMIT..."` can be replaced by `query += " LIMIT..."`

18446744073709551615:
If this is just an arbitrary large number then I would make it a variable called something like that to clarify.

importer/importer.py
```diff
@@ -43,8 +43,17 @@ def __init__(self, countryname, languagename):
...
-def make_query(specific_table):
-    return ('select DISTINCT *  from `{}`').format(specific_table)
+def make_query(specific_table, offset):
+    """
+    Generate a query to retrieve data from database.
+
+    :param specific_table: Name of table to retrieve data from.
+    :param offset: Optional offset to start from.
+    """
+    query = 'select DISTINCT *  from `{}`'.format(specific_table)
```
lokal-profil: Is the double space on purpose?

Vesihiisi: Nope, didn't see it! Out it goes.

importer/importer.py
Vesihiisi: Apparently, in MySQL the only way to use OFFSET is as an argument of LIMIT, and then you have to specify a limit as well. [It's even in the documentation](https://dev.mysql.com/doc/refman/5.7/en/select.html) so, seriously, yeah (ctrl + f 18446744073709551615)

importer/importer.py
```diff
@@ -50,9 +50,10 @@ def make_query(specific_table, offset):
     :param specific_table: Name of table to retrieve data from.
     :param offset: Optional offset to start from.
     """
-    query = 'select DISTINCT *  from `{}`'.format(specific_table)
+    query = 'select DISTINCT * from `{}`'.format(specific_table)
     if isinstance(offset, int):
-        query = query + " LIMIT {}, 18446744073709551615;".format(str(offset))
+        unlimited = "18446744073709551615" # workaround for MySQL requiring a limit when using offset
```
lokal-profil: I think this line might bow be too long (i.e. >79 chars). Also two spaces before comment-`#`.

lokal-profil: HUm.. I would have expected that the CI should pick up the two spaces bit

PR #64, opened 2017-05-20
Austria monuments

Vesihiisi:
Base class for processing Austria in German.

Preview: https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/at_(de)/preview

Task: https://phabricator.wikimedia.org/T166771


importer/AtDe.py
```diff
@@ -0,0 +1,177 @@
...
+        * 251, gegenüber
+          Filtering by "starts with digit" and "contains comma",
+          as we can't rely only on "starts with digit" because
+          10. Oktober-Straße 18 is legit.
+        * zwischen Entensteinstraße 36 u. 38
+          Filtering by "zwischen".
+        * 76
+          Filtering by containing only digits.
+        """
+        bad_address = None
```
lokal-profil: Would it be clearer to use `bad_address = False` as you don't make use of the later contents of `bad_address`?

importer/AtDe.py
```diff
@@ -0,0 +1,177 @@
...
+        if self.has_non_empty_attribute("adresse"):
+            if (utils.first_char_is_number(self.adresse) and
+                    "," in self.adresse):
+                bad_address = self.adresse
+            elif "zwischen" in self.adresse:
+                bad_address = self.adresse
+            elif self.adresse.isdigit():
+                bad_address = self.adresse
+            else:
+                address = self.adresse
```
lokal-profil: Would it be clearer to have this as the `else` for `if bad_address`? Separating the finding-bad-ones logic from the report-or-add logic.

Also not sure `address = self.adresse` is needed since you only use it on the line below.

importer/AtDe.py
```diff
@@ -0,0 +1,177 @@
...
+
+        Using the gemeindekennzahl field, municipality ID,
+        mapped to municipality WD items in external file.
+        """
+        all_codes = self.data_files["municipalities"]
+        if self.has_non_empty_attribute("gemeindekennzahl"):
+            municip_code = str(self.gemeindekennzahl)
+            match = [x["item"]
+                     for x in all_codes if x["value"] == municip_code]
+            if len(match) == 0:
```
lokal-profil: is it safer to do
```python
if len(match) == 1:
  add stuff
else:
   report
```

importer/AtDe.py
```diff
@@ -0,0 +1,177 @@
...
+        """
+        type_match = self.get_type_keyword()
+        municipality_name = self.get_municipality_name()
+        base_desc_german_english = "{} in {}"
+        base_desc_swedish = "{} i {}"
+        place_german = "Österreich"
+        place_english = "Austria"
+        place_swedish = "Österike"
+        type_german = "Denkmalgeschütztes Objekt"
+        type_english = "cultural property"
```
Vesihiisi: Note: rm english and swedish.

PR #65, opened 2017-05-31
Use Property in problem report

Vesihiisi:
Add an option in the report generator to also include the P-id to which the value should be matched.

It's optional so that it can be omitted, e.g. if it's data for a label/description.

Example of report: https://tools.wmflabs.org/coh/_total_se-ship_new.json

Task: https://phabricator.wikimedia.org/T166648


importer/Monument.py
```diff
@@ -487,8 +487,26 @@ def add_to_report(self, key_name, raw_data):
...
+        :type key_name: string
+        :param raw_data: the data that we failed to process
+        :type raw_data: string
+        :param prop_name: name of the property,
+                          as stated in the props library file
+        :type prop_name: string
+        """
+        prop = None
+        if prop_name:
+            try:
```
lokal-profil: Instead of the try/except you could use `prop = self.props.get(prop_name)` which would return `None` if not found.

Vesihiisi:
Re: version bump, it's because we're using the newest pywikibot which does not contain wikidataquery, but the old wikidatastuff still tried to import it, [causing the build to break](https://travis-ci.org/Vesihiisi/COH-tools/jobs/237848059). So it's not relevant to this patch, just to get it through CI.

importer/Monument.py
Vesihiisi: Done.

lokal-profil:
Ah. Explains it. Thanks for letting me know.

PR #66, opened 2017-06-01
Lookup auto

Vesihiisi:
Rewrite of `lookup_table_to_json.py` as a class incorporated into the Monument workflow, so that relevant lookup tables can be fetched automatically in the beginning of the process, rather than downloaded and saved manually.

Implement this in SeFornmin just to make sure it works.

Task: https://phabricator.wikimedia.org/T166790


importer/LookupTable.py
```diff
@@ -0,0 +1,68 @@
...
+        site = pwb.Site("wikidata", "wikidata")
+        return pwb.Page(site, BASE_URL + tablename)
+
+    def get_page_metadata(self, page):
+        """Get permalink and last changed timestamp of wiki page."""
+        timestamp = page.editTime().isoformat()
+        permalink = page.permalink()
+        return {'permalink': permalink, 'timestamp': timestamp}
+
+    def table_to_json(self, tablename):
```
lokal-profil: It would be nice with an example (or test) showing what the source data looks like.

For testing/clarity it might be easier to split this into two functions where one takes `page.get()` as it's input and returns `lookup_dict["mappings"].` Then you could test it easily without needing to fetch a live page.

importer/importer.py
```diff
@@ -106,10 +107,10 @@ def create_connection(arguments):
                                          "data_files": {"counties": "estonia_counties.json"}},
                    "monuments_se-fornmin_(sv)":
                    {"class": SeFornminSv,
-                    "data_files":
-                    {"municipalities": "sweden_municipalities.json",
-                     "socken": "sweden_socken.json",
-                     "types": "se-fornmin_(sv)_types.json"}},
+                             "data_files": {"municipalities": "sweden_municipalities.json",
```
lokal-profil: Indentation looks wrong here. "data_files" and "class" should be on the same level.

importer/LookupTable.py
Vesihiisi: I've split up the thing and separated the functions for fetching a page, extracting a wikitable and converting it into json.

tests/test_LookupTable.py
```diff
@@ -0,0 +1,27 @@
...
+# -*- coding: utf-8  -*-
+import unittest
+import importer.LookupTable as Lt
+
+
+class TestTableConverter(unittest.TestCase):
+
+    def setUp(self):
+        self.converter = Lt.LookupTable("test")
+        self.wikipage = """lorem ipsum\n{| class="wikitable sortable"\n! Name\n! Count\n! Item(s)\n! Notes\n|-\n| kloster\n|\n| {{Q|Q44613}}\n|\n|-\n| pestsäule\n|31\n| {{Q|Q1549521}}\n| \n|}"""
```
lokal-profil: could you line break this on `\n`'s. It would make it easier to read and it would look more like "familiar" wikitext.
A suggestion is to use single quotes but wrap it all in a bracket to keep sane indentation but avoid `\` line endings.

tests/test_LookupTable.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def test_wikipage_to_table(self):
+        output = """{| class="wikitable sortable"\n! Name\n! Count\n! Item(s)\n! Notes\n|-\n| kloster\n|\n| {{Q|Q44613}}\n|\n|-\n| pestsäule\n|31\n| {{Q|Q1549521}}\n| \n|}"""
```
lokal-profil: Also this one

importer/LookupTable.py
```diff
@@ -51,18 +83,28 @@ def table_to_json(self, tablename):
             if len(parsed_wd_items) > 0:
                 dict_key = cells[0].contents.title().strip()
                 for index, item in enumerate(parsed_wd_items):
-                    itemID = item.params[0].title()
-                    if itemID[0] != "Q":
-                        itemID = "Q" + itemID
-                    parsed_wd_items[index] = itemID
-                lookup_dict["mappings"][dict_key] = {
+                    item_id = item.params[0].title()
+                    if item_id[0] != "Q":
```
lokal-profil: user `item_id.startswith("Q")` instead.

tests/test_LookupTable.py
Vesihiisi: Done.

tests/test_LookupTable.py
Vesihiisi: Done.

importer/LookupTable.py
Vesihiisi: done.

lokal-profil:
Ok to merge (from mobile)

PR #69, opened 2017-06-13
BBR duplicate ids patch

Vesihiisi:
Avoid having more than one BBR ID in an item by checking whether the associated WD item already has one that is different from the one in the data, and if so, decoupling it from the data object.

Task: https://phabricator.wikimedia.org/T167860


importer/SeBbrSv.py
```diff
@@ -28,16 +28,19 @@ def update_labels(self):
             self.add_to_report("malformed_label", self.namn)
         self.add_label("sv", label)
 
-    def set_bbr(self):
+    def create_bbr(self):
```
lokal-profil: maybe `create_bbr_link()` to correspond with the name of the created property

importer/SeBbrSv.py
```diff
@@ -109,8 +115,9 @@ def set_heritage_bbr(self):
             except ValueError:
                 self.add_to_report("protection_type", bbr_type)
         kulturarv_source = self.create_kulturarv_source(url)
-        self.add_statement(
-            "heritage_status", type_q, qualifier, refs=[kulturarv_source])
+        if type_q:
```
lokal-profil: Do we want to raise a warning or log if there is no `type_q`?

importer/SeBbrSv.py
```diff
@@ -302,8 +312,44 @@ def exists_with_monument_article(self, language):
...
+
+        Get any P1260 of associated WD item.
+        If it's a BBR ID that's different
+        from the one in the data object,
+        remove the association. Data will be uploaded
+        to a newly created WD item instead.
+        """
+        bbr_url_on_wd = self.get_wditems_bbr_links()
+        if bbr_url_on_wd:
+            for url in bbr_url_on_wd:
```
lokal-profil: you can do this as `if any(url != self.bbr_link for url in bbr_url_on_wd):` the benefit being that it stops as soon as it encounters a hit.

importer/importer_utils.py
```diff
@@ -468,15 +468,24 @@ def get_random_list_sample(some_list, amount):
...
+def get_value_of_property(q_number, property_id, site):
     results = []
     item = pywikibot.ItemPage(site, q_number)
-    if item.exists() and item.claims.get("P31"):
-        for claim in item.claims.get("P31"):
-            results.append(claim.getTarget().getID())
+    if item.exists() and item.claims.get(property_id):
+        for claim in item.claims.get(property_id):
+            target = claim.getTarget()
+            if isinstance(target, str):
```
lokal-profil: I think it might be safer to use
```python
if isinstance(target, pywikibot.ItemPage):
    target = target.getID()
```
Since only `pywikibot.ItemPage` has `getID()`. The current approach would fail if e.g. the value was a WbTime.

(no need for the `if isinstance(target, str): target = target` since it doesn't change anything).

tests/test_importer_utils.py
```diff
@@ -400,6 +400,14 @@ class TestWikidata(unittest.TestCase):
     def setUp(self):
         self.wikidata_site = utils.create_site_instance("wikidata", "wikidata")
 
+    def test_get_value_of_property_string(self):
```
lokal-profil: I would add a test for a non-string value as well.

Vesihiisi: Done.

importer/importer_utils.py

importer/SeBbrSv.py

importer/SeBbrSv.py

Vesihiisi:
In addition to the inline comments I was wondering on the bbra/bbrb relationship.
In the case when we have a building complex (bbra) which only contains one building should we not allow a pre-existing item which contains (only) the bbrb link for that building (and or the bbra link for the complex in question)?

I don't think all links to single buildings use the bbrb prefix.

For example http://kulturarvsdata.se/raa/bbr/21400000696799 is a building.

So in many cases you can't tell just from the url if it's for a building or a complex.

importer/SeBbrSv.py
Vesihiisi: Added this to report.

lokal-profil: Does it happen frequently or is this "just in case"? When/if it happens do we know why?

lokal-profil:
_Replying to [single building discussion](https://github.com/Vesihiisi/COH-tools/pull/69#issuecomment-308657238):_

You are right that not all buildings use the bbrb prefix. In order to do an accurate complex=building matching we would probably need to get some info from the kulturarvsdata.se data.

That said since we already do this for `set_heritage_bbr()` it would probably not be a problem to also get `ns5:hasPart` and use that id for comparisons (if there is only one).

tests/test_importer_utils.py
```diff
@@ -400,6 +400,14 @@ class TestWikidata(unittest.TestCase):
     def setUp(self):
         self.wikidata_site = utils.create_site_instance("wikidata", "wikidata")
 
+    def test_get_value_of_property_page(self):
+        runestone = "Q7899135"
+        country = "P17"
+        sweden = ["Q34"]
+        self.assertEqual(utils.get_value_of_property(runestone,
```
lokal-profil: probably easier to read as:
```python
self.assertEqual(
    utils.get_value_of_property(runestone, country, self.wikidata_site),
    sweden)
```

lokal-profil:
I would then probably split `set_heritage_bbr()` into a "get and store full info for later" method and a "set heritage type method".

importer/SeBbrSv.py
Vesihiisi: IIRC I've only seen it once and made a note about it, but couldn't reproduce it. So it's mostly just in case.

tests/test_importer_utils.py
Vesihiisi: Done.

importer/SeBbrSv.py
```diff
@@ -356,6 +379,9 @@ def __init__(self, db_row_dict, mapping, data_files, existing, repository):
         self.set_monuments_all_id()
         self.set_changed()
         self.wlm_source = self.create_wlm_source(self.monuments_all_id)
+        self.create_bbr_link()
+        self.set_bbr()
+        self.get_online_data()
```
lokal-profil: doesn't this need to come before `create_bbr_link()` to ensure the data has been downloaded.

importer/SeBbrSv.py
```diff
@@ -340,13 +354,22 @@ def check_wd_item(self):
...
+
         If it's a BBR ID that's different
         from the one in the data object,
         remove the association. Data will be uploaded
         to a newly created WD item instead.
         """
         bbr_url_on_wd = self.get_wditems_bbr_links()
+        parts_of_compound = self.get_has_parts()
         if bbr_url_on_wd:
+            if len(parts_of_compound) == 1 and len(bbr_url_on_wd) == 1:
```
lokal-profil: There is one case not handled.
`len(bbr_url_on_wd) == 2` and `len(parts_of_compound) == 1` and the two ids are that of the compound and that of the building.

Easiest structure might be something like the below but I'm open for others:
```python
if len(bbr_url_on_wd) == 0:
    return
elif len(bbr_url_on_wd) == 1:
    if bbr_url_on_wd[0] == self.bbr_link:
        return
    elif len(parts_of_compound) == 1 and parts_of_compound[0] == bbr_url_on_wd[0]
        return
elif len(bbr_url_on_wd) == 2 and len(parts_of_compound) == 1:
    if all(entry in bbr_url_on_wd for entry in [parts_of_compound[0], self.bbr_link]):
        return
self.wd_item["wd-item"] = None
```

importer/SeBbrSv.py
```diff
@@ -28,16 +28,52 @@ def update_labels(self):
...
+        """
+        Retrieve online data about BBR complex.
+
+        This includes, inter alia, the legal protection
+        status and ID's of each part of the
+        compound.
+        """
+        h_property = self.props["cultural_heritage_sweden"]
+        kulturarv_id = self.wd_item["statements"][h_property][0]["value"]
+        url = "http://kulturarvsdata.se/{}".format(kulturarv_id)
```
lokal-profil: could you possibly store the id as `self.kulturarv_id` since you recalculate it once on L129 and once on L138?

importer/SeBbrSv.py
Vesihiisi: Apparently not since when I do print on it it does show the downloaded data.

importer/SeBbrSv.py

importer/SeBbrSv.py
lokal-profil: Sorry. Me mixing things up with `set_heritage_bbr()`

importer/SeBbrSv.py
```diff
@@ -302,11 +329,64 @@ def exists_with_monument_article(self, language):
...
+        keep the item association.
+
+        If it's a BBR ID that's different
+        from the one in the data object,
+        remove the association. Data will be uploaded
+        to a newly created WD item instead.
+        """
+        bbr_url_on_wd = self.get_wditems_bbr_links()
+        parts_of_compound = self.get_has_parts()
+        if len(bbr_url_on_wd) == 1:
```
lokal-profil: How about the case `len(bbr_url_on_wd) == 0`?

lokal-profil: Which I guess can happen if you find a matching wd through the article link but it doens't already have an bbr id claim

Vesihiisi: I think then it shouldn't do anything.

lokal-profil: If I'm not confused (which I might be). When `bbr_url_on_wd` is empty you don't get caught by and of the if/elifs. So you end up at the `self.wd_item["wd-item"] = None`. So the WD connection (if there is one) gets removed even if it should be safe to use it.

lokal-profil:
Ok to merge (sent from mobile so no button)

PR #72, opened 2017-07-28
Move report files to own subdirectory

lokal-profil:
Task: [T171947](https://phabricator.wikimedia.org/T171947)


lokal-profil:
The faling tests all seem to be in files/regions that were not touched. This might be due to flake8 updating (we could pin the version in tox.ini by e.g. `deps = flake8==3.2.1`). That said a bunch of the errors look like stuff we might actually want to fix (e.g. E128, E305, E712, F401, F821, F841, W292, W391)

importer/importer_utils.py
```diff
@@ -34,6 +35,22 @@ def json_to_file(filename, json_content):
...
 
 
+def create_dir(out_path):
+    """
+    Create a directory if it doesn't exist.
+
+    @param out_path: directory to create
+    """
+    if not out_path:
+        raise ValueError('Cannot a create directory without a name.')
```
Vesihiisi: -a

PR #73, opened 2017-08-09
Romania

Vesihiisi:
Mapping: https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/ro_(ro)
Preview: https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/ro_(ro)/preview

Task: https://phabricator.wikimedia.org/T179394


lokal-profil:
I assume this means https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/ro_(ro) is done? Would you mind updating that and the overview table?

lokal-profil:

Vesihiisi:
Preview updated up to here.

importer/RoRo.py
```diff
@@ -1,29 +1,117 @@
...
+        self.add_label("ro", romanian)
+
+    def set_monuments_all_id(self):
+        """Map which column name in specific table to  ID in monuments_all."""
+        self.monuments_all_id = str(self.cod)
+
+    def __init__(self, db_row_dict, mapping, data_files, existing, repository):
+        Monument.__init__(self, db_row_dict, mapping,
+                          data_files, existing, repository)
+        self.set_monuments_all_id()
```
lokal-profil: give "cod" as a parameter and you can drop the local method

importer/RoRo.py
```diff
@@ -1,29 +1,117 @@
...
+                    self.add_statement("location", loc_item)
+
+            if not loc_item:
+                self.add_to_report("localitate", self.localitate, "location")
+
+    def set_wlm_id(self):
+        """
+        Set the WLM ID.
+
+        Do _not_ use the Romania LMI code (P1770)
```
lokal-profil: Could you explain the change from P1770 here? Looks like the other property is already in use on Wikidata so through it it's likely where we would find matching items. If it is a fake identifier than I would assume they all get migrated later on.

importer/RoRo.py
```diff
@@ -1,29 +1,117 @@
...
+        else:
+            self.add_to_report("judetul_iso",
+                               self.judetul_iso,
+                               "located_adm")
 
-    def set_no(self):
-        self.add_statement("romanian_monument_id", self.cod)
+    def set_location(self):
+        """
+        Set Location property from art wlinked in localitate.
```
lokal-profil: typo in this line

importer/RoRo.py
```diff
@@ -1,29 +1,117 @@
...
+                                                      search_term=adm_code,
+                                                      return_content_of="itemLabel",
+                                                      search_in="iso_code")
+        if len(county_item) == 1:
+            place_name = "{}, Romania".format(county_item[0])
+        else:
+            place_name = "Romania"
+        desc = "heritage site in {}".format(place_name)
+        self.add_description("en", desc)
+        self.add_disambiguator(utils.remove_markup(str(self.cod)))
```
lokal-profil: what is the reason for the `utils.remove_markup()` here when you didn't use it in `set_monuments_all_id`?

importer/RoRo.py
```diff
@@ -1,29 +1,117 @@
...
+
+    def set_address(self):
+        if self.has_non_empty_attribute("adresa"):
+            adr_lower = self.adresa.lower()
+            adr_nice = utils.remove_markup(self.adresa)
+            if "piața" in adr_lower or "str." in adr_lower:
+                if self.has_non_empty_attribute("localitate"):
+                    town = utils.remove_markup(self.localitate)
+                    adr_nice = "{}, {}".format(adr_nice,
+                                               town)
```
lokal-profil: no need to linebreak this one

importer/RoRo.py
```diff
@@ -1,29 +1,117 @@
...
-class RoRo(Monument):
 
-    def update_labels(self):
-        name = utils.remove_markup(self.denumire)
-        self.add_label("ro", name)
+class RoRo(Monument):
 
     def set_adm_location(self):
-        return
+        counties = self.data_files["counties"]
```
lokal-profil: You can replace all of this with the `self.set_from_dict_match()` notation

importer/RoRo.py
Vesihiisi: They don't have any external references and according to https://www.wikidata.org/wiki/Wikidata:Property_proposal/Archive/29#P1770 they were originally created to link to rowp.

importer/RoRo.py
Vesihiisi: Done.

importer/RoRo.py
Vesihiisi: No reason. I think it's a leftover from when there was something else inside of this. Removed it.

importer/RoRo.py
Vesihiisi: Done.

importer/RoRo.py
Vesihiisi: Done.

importer/RoRo.py
Vesihiisi: Done.

Vesihiisi:
Updated preview up to here.

importer/RoRo.py
lokal-profil: It is still the one which is de facto used and every P1770 entry should also be in the dataset we are trying to import.

That there is a case for getting rid of the Property  (and migrating them to WLM_ID with a prefix) is something which needs to first be raised on Wikidata.

Vesihiisi: OK. reverting to it.

Vesihiisi:
Preview up to here.

importer/RoRo.py
```diff
@@ -1,29 +1,95 @@
...
+            place_name = "Romania"
+        desc = "heritage site in {}".format(place_name)
+        self.add_description("en", desc)
+        self.add_disambiguator(str(self.cod))
+
+    def set_address(self):
+        if self.has_non_empty_attribute("adresa"):
+            adr_lower = self.adresa.lower()
+            adr_nice = utils.remove_markup(self.adresa)
+            if "piața" in adr_lower or "str." in adr_lower:
```
lokal-profil: I think "Bd." can also be used as a pattern here (based on preview).

maybe break these out as a tuple then:
```python
street_patterns = ("piața", "str.", "br.")
if any(pattern in adr_lower for pattern in street_patterns):
```

Vesihiisi: Added Bd.

Vesihiisi: Updated.

Vesihiisi:
preview up to here

PR #81, opened 2017-08-21
[FOR discussion]Move all dataset/monument type specific information into the class file

lokal-profil:
The first aim for this patch is to introduce a mechanism by which importer.py (and other general classes) do not need to be updated when a new country class is added. This is part of the larger aim of making it clearer how to set up a new country.

I used `SeFornminSv` as an example to allow this to be tested while keeping the changes small. I also tried to ensure that the code is backwards compatible so as to not require all pre-existing classes be updated straight away.

An importer (for one of the new classes) would now instead be triggered by running
`python3 SeFornminSv.py --host localhost --user foo --password bar --db wlm --short --upload live`

Todo:
* Test that this actually works
* Update README to reflect the new entry points
* Set up a task for migrating remaining classes
** Once migrated drop the unnecessary function and command line arguments

Task: [T173693](https://phabricator.wikimedia.org/T173693)


lokal-profil:
I haven't tested that this works as I haven't got a local version of the heritage database on my computer and I don't want to check this out on labs in case it interferes with any ongoing uploads.

Vesihiisi:
I did some minor corrections so that the mapping / data files would be read correctly, and I did [a test upload to the Sandbox which worked](https://www.wikidata.org/w/index.php?title=Q4115189&oldid=543820227), so I'm going to update the documentation now.

lokal-profil:
Not sure about the first commit you did. Isn't `data_files` a property of the `SeFornminSv` since `self.data_files = data_files` in `Monument.__init__`?

Vesihiisi:
I'm not sure why this works either. If we keep it as `self.data_files`, then this happens:

````
USING RANDOM SAMPLE OF 1
Loaded offline data file: mappings/sweden_socken.json
Loaded offline data file: mappings/sweden_municipalities.json
Loaded online data: se-fornmin_(sv)/types
Traceback (most recent call last):
  File "SeFornminSv.py", line 200, in <module>
    importer.main(args, dataset)
  File "/home/alicia/Dropbox/work/COH-tools/importer/importer.py", line 236, in main
    get_items(connection, dataset, upload, short, offset, table)
  File "/home/alicia/Dropbox/work/COH-tools/importer/importer.py", line 187, in get_items
    row, mapping, data_files, existing, wikidata_site)
  File "SeFornminSv.py", line 23, in __init__
    self.set_adm_location()
  File "SeFornminSv.py", line 100, in set_adm_location
    municip_dict = self.data_files["municipalities"]
TypeError: 'NoneType' object is not subscriptable
<class 'TypeError'>
CRITICAL: Closing network session.
````
So the importer loads the data files but the specific class doesn't access them somehow.

lokal-profil:
Hum. Could you check if the `Monument` initialiser gets the right `data_files` data and possibly also if `load_data` returns the right thing?

Technically `SeFornminSv` shouldn't be aware of `dataset` at all.

Vesihiisi:
Looks like line 37 in the Monument initializer is the culprit... I'll remove it and revert the changes in SeFornminSv and see what happens.

Vesihiisi:
It works now even though I'm not sure what I did again.

lokal-profil:
Ah. That was an accidental leftover from my first attempt at rebuilding (before I introduced `Dataset`). You can actually remove the next line as well since that also isn't used.

Did you also try it with one of the other classes?

Vesihiisi:
Yes, I tried it with one of the classes I'm writing now so I'll update that before it goes to review.

PR #82, opened 2017-08-21
Malta

Vesihiisi:
Base class for Malta import.

Mapping: https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/mt_(de)
Preview: https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/mt_(de)/preview

Task: https://phabricator.wikimedia.org/T173926


lokal-profil:

importer/MtDe.py
```diff
@@ -0,0 +1,96 @@
...
+
+        Use the 'gemeinde' field as
+        municipality.
+        """
+        municip_raw = self.gemeinde
+        municip_q = utils.q_from_wikipedia("de", municip_raw)
+        if municip_q:
+            self.add_statement("located_adm", municip_q)
+        else:
+            self.add_to_report("gemeinde", self.gemeinde, "located_adm")
```
lokal-profil: Is it an idea to fall back on `region-iso` for the ones where this fails? (in addition to report)

importer/MtDe.py
```diff
@@ -0,0 +1,96 @@
...
+        self.set_adm_location()
+        self.set_street_address()
+        self.set_wd_item(self.find_matching_wikidata(mapping))
+
+
+if __name__ == "__main__":
+    """Point of entrance for importer."""
+    args = importer.handle_args()
+    dataset = Dataset("mt", "de", MtDe)
+    dataset.data_files = {}
```
lokal-profil: Can drop both this and the next once https://github.com/Vesihiisi/COH-tools/pull/84#discussion_r136830720 has been fixed.

importer/MtDe.py
```diff
@@ -0,0 +1,96 @@
...
+        else:
+            self.add_to_report("gemeinde", self.gemeinde, "located_adm")
+
+    def update_labels(self):
+        labels = {
+            "de": self.name_de,
+            "en": self.name_en,
+            "mt": self.name_mt
+        }
+        for language in labels:
```
lokal-profil:
```python
for lang, label in labels.items():
    self.add_label(lang, label)
```

importer/MtDe.py
```diff
@@ -0,0 +1,96 @@
...
+        do not include the municipality name.
+        So add it to the address after stripping potential
+        bracketed part.
+        """
+        if utils.contains_digit(self.adresse):
+            city_clean = utils.get_rid_of_brackets(self.gemeinde)
+            address = "{}, {}".format(self.adresse, city_clean)
+            self.add_statement("located_street", address)
+
+    def exists_with_monument_article(self, language):
```
lokal-profil: A question which also affects other coutnries. What hapens with a monuments_article of the form "Victoria (Malta)#Zitadelle"? (that is inventarnummer=02 for malta)

The correct behaviour should be that it is interpreted as there being no value at all (since we cannot make a connection between an article section and a Wikidata item).

importer/MtDe.py
```diff
@@ -0,0 +1,96 @@
...
+        if self.has_non_empty_attribute("gemeinde"):
+            municip_base = utils.get_rid_of_brackets(self.gemeinde)
+            placename = "{}, {}".format(municip_base, "Malta")
+        self.add_description("en", description.format(placename))
+
+    def set_monuments_all_id(self):
+        self.monuments_all_id = self.inventarnummer
+
+    def set_heritage_id(self):
+        self.add_statement("malta_monument_id", self.inventarnummer)
```
lokal-profil: the number needs to be zero-padded to 5 digits per [P1799](https://www.wikidata.org/wiki/Property:P1799)

importer/MtDe.py
```diff
@@ -0,0 +1,96 @@
...
+        """
+        Set the descriptions in English.
+
+        Try to add the place name. Since the municipality
+        names are titles of articles, sometimes they
+        have the form of "$municipality (Malta)", in which
+        case remove the bracketed part.
+        """
+        placename = "Malta"
+        description = "cultural property in {}"
```
lokal-profil: I'll reach out and see if I can't get a description template for maltese.

lokal-profil: The maltese pattern would be `propjeta' kulturali ġewwa <municipality>, Malta`

importer/MtDe.py
```diff
@@ -0,0 +1,96 @@
...
+    def exists_with_monument_article(self, language):
+        return super().exists_with_monument_article("de", "artikel")
+
+    def __init__(self, db_row_dict, mapping, data_files, existing, repository):
+        Monument.__init__(self, db_row_dict, mapping,
+                          data_files, existing, repository)
+        self.set_monuments_all_id()
+        self.set_changed()
+        self.wlm_source = self.create_wlm_source(self.monuments_all_id)
+        self.set_country()
```
lokal-profil: You should add `registrar_url` and then use it as a [P973](https://www.wikidata.org/wiki/Property:P973) qualifier on the [P1799](https://www.wikidata.org/wiki/Property:P1799) per https://www.wikidata.org/wiki/Property_talk:P1799#Formater_url.

importer/MtDe.py
Vesihiisi: Done.

importer/MtDe.py
Vesihiisi: Done.

importer/MtDe.py
Vesihiisi: Done.

importer/MtDe.py
Vesihiisi: Yes, and also other will be possible to be removed if Belgium/Brussels is merged before this.

importer/MtDe.py
Vesihiisi: Added a check to the function in Monument so that it ignores titles with #.

importer/MtDe.py
Vesihiisi: Done.

importer/MtDe.py
Vesihiisi: Added a check.

Vesihiisi:
Should not be merged before # 85 in order to implement the general changes in there.

importer/MtDe.py
```diff
@@ -25,30 +33,36 @@ def update_labels(self):
...
+            "mt": "propjeta' kulturali ġewwa {}"
+        }
         placename = "Malta"
-        description = "cultural property in {}"
-        if self.has_non_empty_attribute("gemeinde"):
-            municip_base = utils.get_rid_of_brackets(self.gemeinde)
-            placename = "{}, {}".format(municip_base, "Malta")
-        self.add_description("en", description.format(placename))
+        for lg, descr in patterns.items():
+            if self.has_non_empty_attribute("gemeinde"):
```
lokal-profil: _Note in case you update this anyway after #85 is merged, otherwise you can leave as is._

The if can be moved outside of the loop since placename is the same for both.

importer/MtDe.py
```diff
@@ -9,13 +9,21 @@ def set_adm_location(self):
...
         """
+        adm_q = None
         municip_raw = self.gemeinde
-        municip_q = utils.q_from_wikipedia("de", municip_raw)
-        if municip_q:
-            self.add_statement("located_adm", municip_q)
+        adm_q = utils.q_from_wikipedia("de", municip_raw)
+        if adm_q is None:
+            adm_iso_match = [x for x in self.data_files[
+                "regions"] if x["iso"] == self.region_iso]
```
lokal-profil: you called the file "councils" not "regions"

importer/MtDe.py
Vesihiisi: Yes, as I commented above this will be influenced by #85 so it's waiting for that one.

importer/MtDe.py
Vesihiisi: Done.

PR #83, opened 2017-08-21
Russia

Vesihiisi:


lokal-profil:

importer/RuRu.py
```diff
@@ -0,0 +1,107 @@
...
+
+if __name__ == "__main__":
+    """Point of entrance for importer."""
+    args = importer.handle_args()
+    dataset = Dataset("ru", "ru", RuRu)
+    dataset.data_files = {
+        "administrative": "russia_administrative.json",
+        "subyekty": "russia_subyekty.json"
+    }
+    dataset.lookup_downloads = {}
```
lokal-profil: No need to provide this if it's empty.

importer/RuRu.py
```diff
@@ -0,0 +1,107 @@
...
+        desc = "cultural heritage site in Russia"
+        self.add_description("en", desc)
+
+    def set_heritage_id(self):
+        self.add_statement("kulturnoe-nasledie", self.id)
+
+    def set_monuments_all_id(self):
+        self.monuments_all_id = self.id
+
+    def exists_with_monument_article(self, language):
```
lokal-profil: no need to override this as fieldname is default

importer/RuRu.py
```diff
@@ -0,0 +1,107 @@
...
+            self.add_statement("located_adm", district_match[0]["item"])
+        else:
+            sub_match = [x for x in level_two if x[
+                "value"].lower() == self.region_iso]
+            if len(sub_match) == 1:
+                self.add_statement("located_adm", sub_match[0]["item"])
+            else:
+                self.add_to_report("district", self.district, "located_adm")
+
+    def set_location(self):
```
lokal-profil: With [T174934](https://phabricator.wikimedia.org/T174934) it should be possible to get to the Qid directly from `city_id` if that has been added on Wikivoyage. The current structure can be used as a fallback.

importer/RuRu.py
```diff
@@ -0,0 +1,107 @@
+from Monument import Monument, Dataset
+import importer_utils as utils
+import importer as importer
+
+
+class RuRu(Monument):
+
+    def set_adm_location(self):
```
lokal-profil: With [T174934](https://phabricator.wikimedia.org/T174934) it should be possible to get to the Qid directly from `district_id` if that has been added on Wikivoyage. The current structure can be used as a fallback.

importer/RuRu.py
```diff
@@ -0,0 +1,107 @@
...
+                        return
+                    else:
+                        self.add_to_report("city", self.city, "location")
+
+    def update_labels(self):
+        self.add_label("ru", self.name)
+
+    def update_descriptions(self):
+        desc = "cultural heritage site in Russia"
+        self.add_description("en", desc)
```
lokal-profil: Would be nice to get this in Russian as well. Maybe drop a line on [Wikidata:Форум](https://www.wikidata.org/wiki/Wikidata:%D0%A4%D0%BE%D1%80%D1%83%D0%BC)?

importer/RuRu.py
```diff
@@ -0,0 +1,107 @@
...
+            if city_q_try and city_q_try != administrative_q:
+                city_q_is = utils.get_P31(city_q_try, self.repo)
+                for p31_value in city_q_is:
+                    if p31_value in self.data_files["settlement"]:
+                        self.add_statement("location", city_q_try)
+                        # there can be more than one P31,
+                        # but after first positive
+                        # we can leave
+                        return
+                    else:
```
lokal-profil: I think we want the report on the same level as the loop. Otherwise we will get a report for the first 4 non-settlement P31s even if the fifth is a settlement.

importer/RuRu.py
```diff
@@ -0,0 +1,107 @@
...
+        Of course it's often a village etc. :)
+        Make sure that the target page has a P31 of a
+        settlement, to avoid disambigs etc.
+        Also, compare with 'administrative location' of the
+        object, and don't add if it's the same.
+        """
+        if self.has_non_empty_attribute("city"):
+            city_q_try = utils.q_from_wikipedia("ru", self.city)
+            try:
+                administrative_q = self.wd_item[
```
lokal-profil: would changing this to `admin_q` remove the need for the linebreak?

importer/RuRu.py
```diff
@@ -0,0 +1,107 @@
...
+
+        Preferably try to resolve the 'district' (район/округ,
+        the field contains different types so that's why we're
+        matching it against the whole big list of administrative
+        divisions of different kinds).
+        If that doesn't work, use the higher level субъект.
+        """
+        level_three = self.data_files["administrative"]
+        level_two = self.data_files["subyekty"]
+        district_match = [x for x in level_three if x[
```
lokal-profil:
```python
district_match = [x for x in level_three
                  if x["itemLabel"] == self.district]
```
to avoid line breaking in the middle of the `if`

importer/RuRu.py
```diff
@@ -0,0 +1,107 @@
...
+            "itemLabel"] == self.district]
+        if len(district_match) == 1:
+            self.add_statement("located_adm", district_match[0]["item"])
+        else:
+            sub_match = [x for x in level_two if x[
```
lokal-profil:
```python
sub_match = [x for x in level_two
             if x["value"].lower() == self.region_iso]
```
to avoid line breaking in the `if`

importer/RuRu.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def set_adm_location(self):
+        """
+        Set the administrative location.
+
+        Preferably try to resolve the 'district' (район/округ,
+        the field contains different types so that's why we're
+        matching it against the whole big list of administrative
+        divisions of different kinds).
+        If that doesn't work, use the higher level субъект.
```
lokal-profil: higher __region__ level ... (to connect to field name).

I see that there is both a `region` and a `region_iso` field. IS the second always used or is the former sometimes used instead?

lokal-profil:
I noticed that you use q_from_wikipedia. ous the mechanism aware that this dataset lives in ru.wikivoyage?

PR #84, opened 2017-08-24
Hessen

Vesihiisi:
Mapping: https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/de-he_(de)
Preview: https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/de-he_(de)/preview

Task:

### DIFF
--- a/importer/DeHeDe.py
+++ b/importer/DeHeDe.py
@@ -1,0 +1,29 @@
+from Monument import Monument, Dataset
+import importer_utils as utils
+import importer as importer
+
+
+class DeHeDe(Monument):
+
+    def exists_with_monument_article(self, language):
+        return super().exists_with_monument_article("de", "artikel")
+
+    def __init__(self, db_row_dict, mapping, data_files, existing, repository):
+        Monument.__init__(self, db_row_dict, mapping,
+                          data_files, existing, repository)
+        self.set_monuments_all_id()
+        self.set_changed()
+        self.wlm_source = self.create_wlm_source(self.monuments_all_id)
+        self.set_country()
+        self.set_source()
+        self.set_registrant_url()
+        self.set_wd_item(self.find_matching_wikidata(mapping))
+
+
+if __name__ == "__main__":
+    """Point of entrance for importer."""
+    args = importer.handle_args()
+    dataset = Dataset("de-he", "de", DeHeDe)
+    dataset.data_files = {}
+    dataset.lookup_downloads = {}
+    importer.main(args, dataset)


importer/mappings/de-he_(de).json
```diff
@@ -0,0 +1,21 @@
+{
+    "country": {
+        "item": "Q40",
```
lokal-profil: Q183.

Since the preview is correct I'm wondering if you maybe have local uncommitted changes?

importer/DeHeDe.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def set_location(self):
+        if self.has_non_empty_attribute("ortsteil"):
+            location_q = utils.q_from_wikipedia("de", self.ortsteil)
+            if location_q:
+                self.add_statement("location", location_q)
+
+    def set_adm_location(self):
+        city_q = utils.q_from_wikipedia("de", self.stadt)
+        self.add_statement("located_adm", city_q)
```
lokal-profil: would be good to check that you got a value for `city_q` and if not add to report instead

importer/DeHeDe.py
```diff
@@ -0,0 +1,90 @@
...
+import importer_utils as utils
+import importer as importer
+
+
+class DeHeDe(Monument):
+
+    def set_location(self):
+        if self.has_non_empty_attribute("ortsteil"):
+            location_q = utils.q_from_wikipedia("de", self.ortsteil)
+            if location_q:
```
lokal-profil: else add to report

importer/DeHeDe.py
```diff
@@ -0,0 +1,90 @@
...
+                placename = country
+            desc = templates[lg].format(placename)
+            self.add_description(lg, desc)
+
+    def update_labels(self):
+        if self.has_non_empty_attribute("bezeichnung"):
+            german = utils.remove_markup(self.bezeichnung)
+        elif self.has_non_empty_attribute("adresse"):
+            german = utils.remove_markup(self.adresse)
+        self.add_label("de", german)
```
lokal-profil: is it guaranteed that one or the other always exists?

importer/DeHeDe.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def set_heritage_id(self):
+        self.add_statement("denkXweb", str(self.nummer))
+
+    def set_street_address(self):
+        street_with_no = utils.remove_markup(self.adresse)
+        town = utils.get_rid_of_brackets(self.stadt)
```
lokal-profil: Add a note in the docstring about the brackets

importer/DeHeDe.py
```diff
@@ -0,0 +1,90 @@
...
+            "de": "Deutschland",
+            "en": "Germany",
+            "sv": "Tyskland"
+        }
+        templates = {
+            "en": "Kulturdenkmal in {}",
+            "de": "Kulturdenkmal in {}",
+            "sv": "kulturminnesmärke i {}"
+        }
+        for lg in templates:
```
lokal-profil: use "lang" instead (more common)

importer/DeHeDe.py
```diff
@@ -0,0 +1,90 @@
...
+        return super().exists_with_monument_article("de", "artikel")
+
+    def __init__(self, db_row_dict, mapping, data_files, existing, repository):
+        Monument.__init__(self, db_row_dict, mapping,
+                          data_files, existing, repository)
+        self.set_monuments_all_id()
+        self.set_changed()
+        self.wlm_source = self.create_wlm_source(self.monuments_all_id)
+        self.set_source()
+        self.set_registrant_url()
```
lokal-profil: You add this but never seem to make use of it (per preview)?

importer/DeHeDe.py
```diff
@@ -0,0 +1,90 @@
...
+        self.update_descriptions()
+        self.set_wd_item(self.find_matching_wikidata(mapping))
+
+
+if __name__ == "__main__":
+    """Point of entrance for importer."""
+    args = importer.handle_args()
+    dataset = Dataset("de-he", "de", DeHeDe)
+    dataset.data_files = {}
+    dataset.lookup_downloads = {}
```
lokal-profil: No need to give `lookup_downloads` if it is empty.

importer/DeHeDe.py
```diff
@@ -0,0 +1,90 @@
...
+        self.update_labels()
+        self.update_descriptions()
+        self.set_wd_item(self.find_matching_wikidata(mapping))
+
+
+if __name__ == "__main__":
+    """Point of entrance for importer."""
+    args = importer.handle_args()
+    dataset = Dataset("de-he", "de", DeHeDe)
+    dataset.data_files = {}
```
lokal-profil: I noticed that it is currently not possible to skip providing data_files.
This can be fixed by changing either `self.data_files = None` to `self.data_files = {}` in `Datset.__init__()` or 
`file_dict = dataset.data_files` to `file_dict = dataset.data_files or {}` in `importer.load_data_files()`.

importer/DeHeDe.py
lokal-profil: Per the comment [here](https://github.com/Vesihiisi/COH-tools/pull/82#pullrequestreview-60518441): Would a good general strategy be to:
* Do nothing with the regisrar_url if
** The url is already auto generated by the id property on Wikidata
** The url is invalid (no longer valid or was never valid)
* Add it as a P973 qualifier on either the id claim or the heritage status claim otherwise.
?

importer/DeHeDe.py
lokal-profil: Fixed in #94

importer/DeHeDe.py
Vesihiisi: Done.

importer/DeHeDe.py
Vesihiisi: Done, changed to language.

importer/mappings/de-he_(de).json
Vesihiisi: True, missed adding it to a committ...

importer/DeHeDe.py
Vesihiisi: Done.

importer/DeHeDe.py
Vesihiisi: Well, there's nothing technical to guarantee that items with both these fields empty don't make their way into the database at some point, but at least in my snapshot no, there's none with both fields missing. As for the numbers, there's 8500 (out of 17k) with an empty Bezeichnung, as well as 220 with an empty Adresse.

importer/DeHeDe.py
Vesihiisi: Added this as a qualifier to heritage id.

importer/DeHeDe.py
Vesihiisi: Done.

importer/importer_utils.py
```diff
@@ -98,7 +98,6 @@ def remove_markup(text):
     if "[" in text:
         text = wparser.parse(text)
         text = text.strip_code()
-    text = text.replace("&nbsp;", " ")
```
lokal-profil: Looks like this was removed again (and is responsible for the failing test)

importer/DeHeDe.py
```diff
@@ -10,10 +10,16 @@ def set_location(self):
             location_q = utils.q_from_wikipedia("de", self.ortsteil)
             if location_q:
                 self.add_statement("location", location_q)
+            else:
+                self.add_to_report("ortsteil", self.ortsteil, "location")
 
     def set_adm_location(self):
+        city_q = None
```
lokal-profil: Don't think you need to set this since `utils.q_from_wikipedia` returns `None` on failiure

importer/DeHeDe.py
lokal-profil: Works for me. Thanks for checking.

importer/DeHeDe.py
```diff
@@ -45,24 +51,33 @@ def update_labels(self):
         self.add_label("de", german)
 
     def set_heritage_id(self):
-        self.add_statement("denkXweb", str(self.nummer))
+        described_qual = {"described_at_url": self.registrant_url}
```
lokal-profil: Not needed in this case I would say since the Property (on Wikidata) autogenerates the url. (See follow-up comment below)

importer/DeHeDe.py
lokal-profil: For Hessen it seems as though the rl is already auto generated by the id property on Wikidata. Which would fall in the "do nothing" category.

Do you think the decision method above works in general? IS there somewhere else where we would want to use the url?

importer/importer_utils.py
lokal-profil: Fixed in commit during review

importer/DeHeDe.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def exists_with_monument_article(self, language):
+        return super().exists_with_monument_article("de", "artikel")
+
+    def __init__(self, db_row_dict, mapping, data_files, existing, repository):
+        Monument.__init__(self, db_row_dict, mapping,
+                          data_files, existing, repository)
+        self.set_monuments_all_id("nummer")
+        self.set_changed()
+        self.wlm_source = self.create_wlm_source(self.monuments_all_id)
```
lokal-profil: you had a new function for this

Vesihiisi: Done.

Vesihiisi:
Could you specify what the comment "Works for me. Thanks for checking." refers to? Looks like I can't reply directly under it.

importer/DeHeDe.py
Vesihiisi: Removed it.

importer/DeHeDe.py
Vesihiisi: Removed it.

lokal-profil:
@Vesihiisi 

It was a reply to
>Well, there's nothing technical to guarantee that items with both these fields empty don't make their way into the database at some point, but at least in my snapshot no, there's none with both fields missing. As for the numbers, there's 8500 (out of 17k) with an empty Bezeichnung, as well as 220 with an empty Adresse.

PR #85, opened 2017-08-24
Belgium/Brussels

Vesihiisi:
Mapping: https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/be-bru_(nl)
Preview: https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/be-bru_(nl)/preview

Task: https://phabricator.wikimedia.org/T174019


importer/BeBruNl.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def set_address(self):
+        street_with_no = utils.remove_markup(self.adres.title()).rstrip(',')
+        whole_address = "{}, {}".format(street_with_no, self.plaats)
+        self.add_statement("located_street", whole_address)
+
+    def set_special_is(self):
+        objtype = self.objtype.lower()
+        types = self.data_files["type"]["mappings"]
+        special_types = [types[x]["items"]
```
lokal-profil: you use this mechanism so often (also in other classes) that it might be worth introudicing the following in your utils
```python
def get_matching_items_from_dict(value, dict_name):
    """
    Return all items in a dict for which the label matches the provided value.
    
    @param value: the value to match
    @param dict_name: the dict to look in
    """
    return [dict_name[x]["items"]
            for x in styles if x.lower() == value]
```

importer/BeBruNl.py
```diff
@@ -0,0 +1,131 @@
...
+        self.set_heritage()
+        self.set_country()
+        self.set_image()
+        self.set_adm_location()
+        self.set_commonscat()
+        self.set_architect()
+        self.set_style()
+        self.set_is()
+        self.set_special_is()
+        self.set_coords(("lat", "lon"))
```
lokal-profil: I would recomment modifying `Monument.set_coords()` to allow `coord_keywords_tuple` to be `None` then add `coord_keywords_tuple = coord_keywords_tuple or ("lat", "lon")` inside it since that is the dominating tuple (used for all newer mappings).

Then the call here would just be `self.set_coords()`

importer/BeBruNl.py
```diff
@@ -0,0 +1,131 @@
...
+    def set_architect(self):
+        if self.has_non_empty_attribute("bouwdoor"):
+            if utils.count_wikilinks(self.bouwdoor) == 1:
+                arch_q = utils.q_from_first_wikilink("nl", self.bouwdoor)
+                self.add_statement("architect", arch_q)
+            else:
+                self.add_to_report("bouwdoor", self.bouwdoor, "architect")
+
+    def update_labels(self):
+        dutch = utils.remove_markup(self.omschrijving)
```
lokal-profil: Could you add the following test to test_importer_utils:TestWikitext to show that pipes will be handled correctly here.
```python


    def test_remove_markup_pipe(self):
        text = "[[Tegera Arena|Arenan]], huvudentrén"
        output = "Arenan, huvudentrén"
        self.assertEqual(utils.remove_markup(text), output)
```

importer/BeBruNl.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+class BeBruNl(Monument):
+
+    def set_adm_location(self):
+        brussel_region = "Q240"
+        location = self.plaats
+        municip_dict = self.data_files["municipalities"]
+        if location:
+            location_match = [x for x in municip_dict if x[
```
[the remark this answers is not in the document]
Vesihiisi: Done.

importer/BeBruNl.py
```diff
@@ -0,0 +1,131 @@
...
+                "itemLabel"] == location]
+            if len(location_match) == 1:
+                self.add_statement("located_adm", location_match[0]["item"])
+        else:
+            self.add_statement("located_adm", brussel_region)
+
+    def set_heritage(self):
```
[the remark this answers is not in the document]
Vesihiisi: Done.

importer/BeBruNl.py
Vesihiisi: Done and it passes.

importer/BeBruNl.py
Vesihiisi: Done, added this to utils and substituted in class.

importer/BeBruNl.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def set_building_year(self):
+        if self.has_non_empty_attribute("bouwjaar"):
+            if utils.legit_year(self.bouwjaar):
+                self.add_statement(
+                    "inception", {"time_value": {"year": self.bouwjaar}})
+            else:
+                self.add_to_report("bouwjaar", self.bouwjaar, "inception")
+
+    def set_monuments_all_id(self):
```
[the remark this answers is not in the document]
Vesihiisi: Done.

importer/BeBruNl.py
```diff
@@ -0,0 +1,131 @@
...
+        self.monuments_all_id = self.code
+
+    def set_heritage_id(self):
+        self.add_statement("heritage_brussels", self.code)
+
+    def exists_with_monument_article(self, language):
```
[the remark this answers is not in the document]
Vesihiisi: Done.

importer/BeBruNl.py
```diff
@@ -0,0 +1,131 @@
...
+        return super().exists_with_monument_article("nl", "monument_article")
+
+    def __init__(self, db_row_dict, mapping, data_files, existing, repository):
+        Monument.__init__(self, db_row_dict, mapping,
+                          data_files, existing, repository)
+        self.set_monuments_all_id()
+        self.set_changed()
+        self.wlm_source = self.create_wlm_source(self.monuments_all_id)
```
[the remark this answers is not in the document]
Vesihiisi: Done.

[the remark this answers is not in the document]
Vesihiisi: Done.

importer/BeBruNl.py
Vesihiisi: Done.

importer/importer_utils.py
```diff
@@ -191,6 +191,17 @@ def q_from_first_wikilink(language, text):
...
 
 
+def get_matching_items_from_dict(value, dict_name):
+    """
+    Return all items in a dict for which the label matches the provided value.
+
+    @param value: the value to match
+    @param dict_name: the dict to look in
+    """
+    return [dict_name[x]["items"]
```
lokal-profil: is the list comprehension only used in case of a mixture of upper and lower case values?

If so I would collapse it to one list using (but with good variable names):
```python
ll = [dict_name[x]["items"] for x in dict_name if x.lower() == value]
l2 = []
for l in ll:
 l2 += l
return list(set(l2))
```
that way you have only one list (as needed) rather than a nested list.

importer/BeBruNl.py
```diff
@@ -10,14 +10,23 @@ def set_adm_location(self):
         location = self.plaats
         municip_dict = self.data_files["municipalities"]
         if location:
-            location_match = [x for x in municip_dict if x[
-                "itemLabel"] == location]
+            location_match = [x for x in municip_dict
```
lokal-profil: Maybe als obreak this out as a util function since the basic structure holds for all your data files.

```python
def get_matching_items_from_data_file(data_list, value, value_field="itemLabel", target_field="item"):
    return [x[target_field] for x in data_list if x[value_field] == value]
```

importer/Monument.py
```diff
@@ -255,13 +255,14 @@ def set_heritage(self):
...
 
-    def set_coords(self, coord_keywords_tuple):
+    def set_coords(self, coord_keywords_tuple=None):
         """
         Add coordinates.
 
-        :param coord_keywords_tuple: names of the columns with
-            coordinates, like ("lat", "long")
+        :param coord_keywords_tuple: optionally, names of the columns with
+            coordinates -- by default ("lat", "long")
```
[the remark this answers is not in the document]
Vesihiisi: Done.

importer/importer_utils.py
Vesihiisi: No, there's no check for whether there's a mix... The .lower() is just to make sure if there's an accidental mixture, but there's no rule.

importer/BeBruNl.py
Vesihiisi: Done, added to utils w/tests.

importer/importer_utils.py
lokal-profil: Then I would suggest adding on a logic step which returns the 0th element (or an empty list) so that the result is a list rather than a list of lists.

importer/BeBruNl.py
lokal-profil: See question in chat

lokal-profil:
Might be worth fixing https://github.com/Vesihiisi/COH-tools/pull/84#discussion_r136830720 as part of this PR as well. That way all the dependencies for the other PRs are handled in one place (so they can be rebased or have this merged into them).

lokal-profil:
The comment above about #84 was instead solved by a mini PR at #94

importer/importer_utils.py
Vesihiisi: Ok.

PR #89, opened 2017-08-30
Spain

Vesihiisi:
Mapping: https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/es_(es)
Preview: https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/es_(es)/preview

Task: https://phabricator.wikimedia.org/T174903


importer/importer_utils.py
```diff
@@ -95,7 +95,7 @@ def remove_markup(text):
     remove_br = re.compile('<br.*?>\W*', re.I)
     text = remove_br.sub(' ', text)
     text = " ".join(text.split())
-    if "[" in text:
+    if "[" in text or "'''" in text or "''" in text:
```
lokal-profil: don't need to check for  `'''` as this is caught by `''`

importer/EsEs.py
```diff
@@ -0,0 +1,135 @@
...
+                self.add_to_report("lugar", self.lugar, "location")
+
+    def set_heritage_id(self):
+        self.add_statement("bien_de_interes", str(self.bic))
+
+    def set_monuments_all_id(self):
+        """Map which column name in specific table to  ID in monuments_all."""
+        self.monuments_all_id = self.bic
+
+    def exists_with_monument_article(self, language):
```
lokal-profil: not needed

importer/EsEs.py
```diff
@@ -0,0 +1,135 @@
...
+                special_types = special_match[0]["items"]
+                for s in special_types:
+                    self.add_statement("is", s)
+
+    def update_labels(self):
+        spanish = utils.remove_markup(self.nombre)
+        self.add_label("es", spanish)
+
+    def update_descriptions(self):
+        desc_dic = {"es": "Bien de Interés Cultural",
```
lokal-profil: Could we attempt to add some geo info here?

importer/EsEs.py
```diff
@@ -0,0 +1,135 @@
...
+    def set_special_is(self):
+        is_dic = self.data_files["tipobic"]["mappings"]
+        if self.has_non_empty_attribute("tipobic"):
+            special_raw = self.tipobic.lower()
+            special_match = [is_dic[x]
+                             for x in is_dic if x.lower() == special_raw]
+            if len(special_match) == 1:
+                self.remove_statement("is")
+                special_types = special_match[0]["items"]
+                for s in special_types:
```
lokal-profil: Is it possible to have more than one type in tipobic? If not I don't think it should be possible with multiple matches.

importer/EsEs.py
```diff
@@ -0,0 +1,135 @@
...
+        else:
+            self.add_to_report("municipio", self.municipio, "located_adm")
+
+    def set_special_is(self):
+        is_dic = self.data_files["tipobic"]["mappings"]
+        if self.has_non_empty_attribute("tipobic"):
+            special_raw = self.tipobic.lower()
+            special_match = [is_dic[x]
+                             for x in is_dic if x.lower() == special_raw]
+            if len(special_match) == 1:
```
lokal-profil: add report on no match

importer/EsEs.py
```diff
@@ -0,0 +1,135 @@
...
+        spanish = utils.remove_markup(self.nombre)
+        self.add_label("es", spanish)
+
+    def update_descriptions(self):
+        desc_dic = {"es": "Bien de Interés Cultural",
+                    "en": "cultural property in Spain"}
+        for lg in desc_dic:
+            self.add_description(lg, desc_dic[lg])
+
+    def set_location(self):
```
lokal-profil: add a docstring explaining that htis is run after adm and we check if it is more specific.

importer/EsEs.py
```diff
@@ -0,0 +1,135 @@
...
+        if self.has_non_empty_attribute("lugar"):
+            if utils.count_wikilinks(self.lugar) == 1:
+                adm_loc = self.wd_item["statements"]["P131"][0]["value"]
```
lokal-profil: If I remember correctly one of the other PRs introduces a function to do this.

importer/EsEs.py
```diff
@@ -0,0 +1,135 @@
...
+        municip_dic = self.data_files["municipalities"]
+        province_dic = self.data_files["provinces"]
+        if utils.count_wikilinks(self.municipio) == 1:
+            adm_q = utils.q_from_first_wikilink("es", self.municipio)
+        else:
+            # sometimes they're in ''
+            municip_raw = utils.remove_markup(self.municipio)
+            municip_match = [x for x in municip_dic if x[
+                "itemLabel"] == municip_raw]
+            if len(municip_match) == 1:
```
lokal-profil: set muni report as an else here instead. We want it reported even if we fall back on a province

importer/EsEs.py
```diff
@@ -0,0 +1,135 @@
...
+                adm_q = municip_match[0]["item"]
+        if not adm_q:
+            province_raw = self.provincia_iso.lower()
+            prov_match = [x for x in province_dic if x[
+                "itemLabel"].lower() == province_raw]
+            if len(prov_match) == 1:
```
lokal-profil: add a province report as an else here, just in case

importer/EsEs.py
```diff
@@ -0,0 +1,135 @@
...
+        self.add_statement("bien_de_interes", str(self.bic))
+
+    def set_monuments_all_id(self):
+        """Map which column name in specific table to  ID in monuments_all."""
+        self.monuments_all_id = self.bic
+
+    def exists_with_monument_article(self, language):
+        return super().exists_with_monument_article("es", "monument_article")
+
+    def __init__(self, db_row_dict, mapping, data_files, existing, repository):
```
lokal-profil: It looks like the monuments database defaults to `bic="0"` when no value is given. So we probably want to filter out all of those. I also spotted one with `bic="-"`.

Maybe filter out all with `bic in ("", "0", "-")`

lokal-profil:
Since I just tried it for the Iranian dataset: try `dateparser.parse(spanish_date, languages=['es'])` which should remove the need for an explicit parser here.

Vesihiisi:
Moved date handling to dateparser.

importer/importer_utils.py
Vesihiisi: Removed.

importer/EsEs.py
Vesihiisi: Removed.

importer/EsEs.py
Vesihiisi: Done.

importer/EsEs.py
Vesihiisi: Theoretically it can happen with lookup tables (you can have two q-items for a certain string). It doesn't seem to happen for this dataset, so I've removed it.

importer/EsEs.py
Vesihiisi: Done.

importer/EsEs.py
Vesihiisi: Added municip or lugar to the English description.

importer/EsEs.py
Vesihiisi: Done.

importer/EsEs.py
Vesihiisi: Which one do you mean? I don't think we've even done this exact thing before.

importer/EsEs.py
Vesihiisi: Added a check so that those are not uploaded. Also added "s/n" to the list.

importer/EsEs.py
Vesihiisi: Done.

Vesihiisi:
Preview up to date with last edits.

importer/EsEs.py
```diff
@@ -1,38 +1,29 @@
...
-                      "abril", "mayo", "junio",
-                      "julio", "agosto",
-                      "septiembre", "octubre",
-                      "noviembre", "diciembre", ]
-            date_parts = self.fecha.split(" ")
-            if len(date_parts) != 5:
+            es_date = dateparser.parse(self.fecha, languages=['es'])
+            try:
+                date_dict = utils.datetime_object_to_dict(es_date)
+            except ValueError:
```
lokal-profil: add `, start_time` as the third argument to "add_to_report" to tie it to a property?

importer/EsEs.py
```diff
@@ -45,45 +36,70 @@ def set_adm_location(self):
...
+        place = "Spain"
+        if self.has_non_empty_attribute("municipio"):
+            place = "{}, {}".format(utils.remove_markup(self.municipio), place)
+        elif self.has_non_empty_attribute("lugar"):
+            place = "{}, {}".format(utils.remove_markup(self.lugar), place)
         desc_dic = {"es": "Bien de Interés Cultural",
-                    "en": "cultural property in Spain"}
+                    "en": "cultural property in {}".format(place)}
         for lg in desc_dic:
+            print(desc_dic[lg])
```
lokal-profil: leftover from testing

importer/EsEs.py
lokal-profil: True. We have a function for adding, removing and substituting statements but not for retrieving the values.

Think it would be useful to add it to Monuments.py and then call it here with `self.get_statement_values("located_adm")` that way we keep a clear separation between Property labels and property ids.

```python
    def get_statement_values(self, prop_name):
        """
        Retrieve all statements with a given property from the data object.

        :param prop_name: name of the property,
            as stated in the props library file
        """
        base = self.wd_item["statements"]
        prop = self.props[prop_name]
        if prop in base:
            return [x['value'] for x in base[prop]]
```

importer/EsEs.py
Vesihiisi: removed!

importer/EsEs.py
Vesihiisi: Added!

Vesihiisi:
Added `Monument.get_statement_values()` since apparently i can't reply to the comment.

PR #91, opened 2017-08-31
Netherlands Aruba

Vesihiisi:
Mapping: https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/nl-aw_(nl)
Preview: https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/nl-aw_(nl)/preview

Task: https://phabricator.wikimedia.org/T174901


importer/NlAwNl.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def set_inception(self):
+        if self.has_non_empty_attribute("bouwjaar"):
+            if utils.legit_year(self.bouwjaar):
+                inc_year = {"time_value": {"year": self.bouwjaar}}
+                self.add_statement("inception", inc_year)
+            else:
+                self.add_to_report("bouwjaar", self.bouwjaar, "inception")
+
+    def set_heritage_id(self):
```
lokal-profil: I'm unsure if we might actually be able to use `AW-` as a prefix here since it is the official iso for the country (within the other country)

importer/NlAwNl.py
```diff
@@ -0,0 +1,93 @@
...
+        country = self.mapping["country_code"].upper()
+        wlm = "{}-{}".format(country, str(self.objectnr))
+        self.add_statement("wlm_id", wlm)
+
+    def set_monuments_all_id(self):
+        """Map which column name in specific table to  ID in monuments_all."""
+        self.monuments_all_id = self.objectnr
+
+    def exists_with_monument_article(self, language):
```
lokal-profil: can be dropped

importer/NlAwNl.py
```diff
@@ -0,0 +1,93 @@
...
+        loc_dic = self.data_files["settlements"]
+        if self.has_non_empty_attribute("plaats"):
+            if utils.count_wikilinks(self.plaats) == 1:
+                loc_q = utils.q_from_first_wikilink("nl", self.plaats)
+            else:
+                loc_match = [x for x in loc_dic if x[
+                    "itemLabel"] == self.plaats]
+                if len(loc_match) == 1:
+                    loc_q = loc_match[0]["item"]
+        if loc_q:
```
lokal-profil: indent this if/else once so that it's only triggered on non-empty plaats

importer/NlAwNl.py
Vesihiisi: I don't actually think it can be completely left out, since there's no way for the master function in Monument to know which language to search in?

importer/NlAwNl.py
Vesihiisi: Done.

importer/NlAwNl.py
Vesihiisi: So that's either NL for the real country, AW for the "country" or NL-AW for the database prefix. Personally I think it either the first or the last one because they're easy to explain, as opposed to "item with a ISO 3166-1 alpha-2 code which may or may not be a sovereign country".

importer/NlAwNl.py
Vesihiisi: But also updated the things related to the recent Monument updates etc

importer/NlAwNl.py
lokal-profil: After looking at the nl-gem case I think the lat inte id's best since it is within the wlm dataset that uniqueness is guaranteed.

Vesihiisi: Changed to WLM prefix. Will also affect other datasets where table name != country iso.

importer/NlAwNl.py
```diff
@@ -33,14 +33,18 @@ def set_location(self):
...
+                                                            search_in="itemLabel",
+                                                            return_content_of="item")
                 if len(loc_match) == 1:
-                    loc_q = loc_match[0]["item"]
-        if loc_q:
-            self.add_statement("location", loc_q)
-        else:
-            self.add_to_report("plaats", self.plaats, "location")
+                    loc_q = loc_match[0]
+                    print(loc_q)
```
lokal-profil: left over from testing

PR #95, opened 2017-09-06
Don't get disambiguations when retrieving Q via WP

Vesihiisi:
When we get wikidata items via wikipedia pages we already ensure that the page isn't a redirect. In addition to this we should also ensure that disambiguation pages aren't returned.

Task: https://phabricator.wikimedia.org/T175070


importer/importer_utils.py
```diff
@@ -171,7 +171,7 @@ def q_from_wikipedia(language, page_title):
     summary = summary.format(page_title, language)
     wd_repo = create_site_instance("wikidata", "wikidata")
     wdstuff = wds(wd_repo, edit_summary=summary)
-    if page.exists():
+    if page.exists() and not page.isDisambig():
```
lokal-profil: Ideally this should be after the redirect resolution since that might also be a disambig

Vesihiisi: Done.

PR #98, opened 2017-09-09
Iran

lokal-profil:
Migration base for Iran

Mapping: [Wikidata:WikiProject_WLM/Mapping_tables/ir_(fa)](https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/ir_(fa))
Preview: [Wikidata:WikiProject_WLM/Mapping_tables/ir_(fa)/preview](https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/ir_(fa)/preview)

Task: [T175420](https://phabricator.wikimedia.org/T175420)


lokal-profil:
Handing this over to @Vesihiisi if that works with you.
To do:
* Generate preview
* Check [T174966](https://phabricator.wikimedia.org/T174966) to see if there is an update about street address.
* Check [T174966](https://phabricator.wikimedia.org/T174966) to see if there is an update about using name also for `q_from_wikipedia()`.

Vesihiisi:
Current state of address info is _in many cases, they are statements such as n km in highway x, for example_, so we're investigating whether those can be told apart from real ones. Otherwise the whole thing will be dropped.

Vesihiisi:
Update: using Directions (P2795) when the address is not a good street address.

lokal-profil:
last one included a rebase

Vesihiisi:
[Preview updated](https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/ir_(fa)/preview)

lokal-profil:
Short update: I managed to generate a [matches preview](https://tools.wmflabs.org/coh/matches_ir_fa.txt)  but unfortunately the text is to large to save on a single Wikidata page.

lokal-profil:
put together a matching list at [Mapping_tables/ir_(fa)/matches](https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/ir_(fa)/matches).

So far I've identified at least "religion" which should be added to the blacklist. I'll ping in [T174966](https://phabricator.wikimedia.org/T174966) to see if we can get help finding others.

Once this is done I think we are ready to merge and put up a bot request.

Vesihiisi:
Other blacklistable items in matches list:
Überrest (Q333109)
symbol (Q80071)
tourist attraction (Q570116) -- it feels like a label, not a state of being
city (Q515) -- ??? not sure how reasonable this one is
bathroom (Q190771) -- ok, looks like a bunch of these already exist -- https://www.wikidata.org/wiki/Q5877871 -- but it looks very wrong. It's _probably_ supposed to mean something like public bath, but if we don't know that for sure, blacklist the Q5877871 because it doesn't fit either way

lokal-profil:
I'm unsure about "tourist attraction (Q570116)". I agree it feels more like a label but I'm assuming that it might get used as a second P31 label? So "house" & "tourist attraction".

Bathroom should probably be [Q28077](https://www.wikidata.org/wiki//Q28077). I agree with blacklisting but we should probably ping people about getting these changed before we run.

PR #99, opened 2017-09-11
Hong Kong

Vesihiisi:
Mapping:
Preview:

Task:

### DIFF
--- a/importer/HkEn.py
+++ b/importer/HkEn.py
@@ -1,0 +1,26 @@
+from Monument import Monument, Dataset
+import importer_utils as utils
+import importer as importer
+
+
+class HkEn(Monument):
+
+    def exists_with_monument_article(self, language):
+        return super().exists_with_monument_article("en", "monument_article")
+
+    def __init__(self, db_row_dict, mapping, data_files, existing, repository):
+        Monument.__init__(self, db_row_dict, mapping,
+                          data_files, existing, repository)
+        self.set_monuments_all_id("id")
+        self.set_registrant_url()
+        self.set_changed()
+        self.set_wlm_source()
+        self.set_country()
+        self.set_wd_item(self.find_matching_wikidata(mapping))
+
+
+if __name__ == "__main__":
+    """Point of entrance for importer."""
+    args = importer.handle_args()
+    dataset = Dataset("hk", "en", HkEn)
+    importer.main(args, dataset)
--- a/importer/mappings/hk_(en).json
+++ b/importer/mappings/hk_(en).json
@@ -1,0 +1,21 @@
+{
+    "country": {
+        "item": "Q233",
+        "name": "malta"
+    },
+    "country_code": "mt",
+    "default_is": {
+        "item": "Q2065736",
+        "name": "cultural property"
+    },
+    "heritage": {
+        "item": "Q1939635",
+        "name": "declared monument"
+    },
+    "language": "en",
+    "table_name": "hk",
+    "unique": {
+        "comment": "WLM ID",
+        "property": "P2186"
+    }
+}
--- a/importer/mappings/hk_(en).json
+++ b/importer/mappings/hk_(en).json
@@ -1,9 +1,9 @@
 {
     "country": {
-        "item": "Q233",
-        "name": "malta"
+        "item": "Q8646",
+        "name": "Hongkong"
     },
-    "country_code": "mt",
+    "country_code": "hk",
     "default_is": {
         "item": "Q2065736",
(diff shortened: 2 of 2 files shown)


importer/HkEn.py
```diff
@@ -0,0 +1,80 @@
...
+            if parsed_d:
+                date_dict = utils.datetime_object_to_dict(parsed_d)
+                qualifier = {"start_time": {"time_value": date_dict}}
+
+        heritage = self.mapping["heritage"]["item"]
+        self.add_statement("heritage_status", heritage, qualifier)
+
+    def set_admin_location(self):
+        """Admin location same for all, no subdivisions."""
+        hkong = "Q8646"
```
lokal-profil: Would it be possible to make use of `region` to add some type of extra to this?

While "Hong Kong Island" (Q19483), "Kowloon" (Q1210635, actually Kowloon Peninsula)  and "New Territories" (Q596660) are just historical subdivisions it's still a guide to figuring out which modern district it corresponds to.

My suggestion is either admin_location with an end_date (1999) or as a qualifier on the hkong value.

importer/HkEn.py
```diff
@@ -0,0 +1,80 @@
...
+        self.add_description("en", english)
+
+    def exists_with_monument_article(self, language):
+        return super().exists_with_monument_article("en", "monument_article")
+
+    def __init__(self, db_row_dict, mapping, data_files, existing, repository):
+        Monument.__init__(self, db_row_dict, mapping,
+                          data_files, existing, repository)
+        self.set_monuments_all_id("id")
+        self.set_registrant_url()
```
lokal-profil: can this be skipped since it's not in the db?

importer/HkEn.py
```diff
@@ -0,0 +1,80 @@
...
+        If declaration_date present and parseable,
+        add it as start_time qualifier
+        """
+        qualifier = None
+        if self.has_non_empty_attribute("declaration_date"):
+            parsed_d = dateparser.parse(self.declaration_date)
+            # will give None when failure
+            if parsed_d:
+                date_dict = utils.datetime_object_to_dict(parsed_d)
+                qualifier = {"start_time": {"time_value": date_dict}}
```
lokal-profil: else add to report

importer/HkEn.py
```diff
@@ -0,0 +1,80 @@
...
+        """Admin location same for all, no subdivisions."""
+        hkong = "Q8646"
+        self.add_statement("located_adm", hkong)
+
+    def set_location(self):
+        loc_q = None
+        if utils.count_wikilinks(self.location) == 1:
+            loc_q = utils.q_from_first_wikilink("en", self.location)
+
+        if loc_q:
```
lokal-profil: else add to report

importer/HkEn.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def update_labels(self):
+        english = utils.remove_markup(self.name)
+        self.add_label("en", english)
+
+    def update_descriptions(self):
+        english = "declared monument of Hong Kong"
+        self.add_description("en", english)
+
+    def exists_with_monument_article(self, language):
```
lokal-profil: I believe this can be skipped since it's the mapped language and the default keyword (L476 in `Monuments.py`)

importer/HkEn.py
```diff
@@ -0,0 +1,80 @@
...
+        self.add_statement("heritage_status", heritage, qualifier)
+
+    def set_admin_location(self):
+        """Admin location same for all, no subdivisions."""
+        hkong = "Q8646"
+        self.add_statement("located_adm", hkong)
+
+    def set_location(self):
+        loc_q = None
+        if utils.count_wikilinks(self.location) == 1:
```
lokal-profil: It seems to be quite common with multiple links where one of them is the link to a district ([query for all 18](https://query.wikidata.org/#%23Cats%0ASELECT%20%3Fitem%20%3FitemLabel%20%0AWHERE%20%0A{%0A%20%3Fitem%20wdt%3AP31%20wd%3AQ50256.%0A%20%3Fitem%20wdt%3AP131*%20wd%3AQ8646.%0A%20SERVICE%20wikibase%3Alabel%20{%20bd%3AserviceParam%20wikibase%3Alanguage%20%22en%22.%20}%0A})).

I would suggest checking if any of the links correspond to one of these. If so add that as admin_location and remove it from the list of detected links. if only one remains then use that for location.

If this is called before set_admin_location then that can quickly detect if a value has been added and only go through with the defaults otherwise.

PR #101, opened 2017-09-12
Tunisia

Vesihiisi:


tests/test_importer_utils.py
```diff
@@ -303,6 +303,16 @@ def test_dict_to_iso_date_year_month(self):
 
 class TestWikitext(unittest.TestCase):
 
+    def test_get_urls_none(self):
+        text = "[[Tegera Arena]], huvudentrén"
+        output = []
+        self.assertEqual(utils.get_external_links(text), output)
+
+    def test_get_urls_one(self):
```
lokal-profil: add one more small test with something like "prefix [url] postfix" i.e. a url which would look like [1] on a wiki. And amend this one to also have a pre- and postfix.

importer/TnFr.py
```diff
@@ -0,0 +1,138 @@
...
+
+        Use 'decret' as 'described at url'.
+        Extract date from 'date' as 'start time'.
+        """
+        quals = {}
+        heritage_item = self.mapping["heritage"]["item"]
+
+        urls = utils.get_external_links(self.decret)
+        if len(urls) == 1:
+            quals["described_at_url"] = urls[0].strip_code()
```
lokal-profil: elif len(urls) > 1:
   add to report

importer/TnFr.py
```diff
@@ -0,0 +1,138 @@
...
+
+        date_templates = utils.wparser.parse(self.date).filter_templates()
+        if len(date_templates) == 1:
```
lokal-profil: elif len(urls) > 1:
add to report

importer/TnFr.py
```diff
@@ -0,0 +1,138 @@
...
+        Extract date from 'date' as 'start time'.
+        """
+        quals = {}
+        heritage_item = self.mapping["heritage"]["item"]
+
+        urls = utils.get_external_links(self.decret)
+        if len(urls) == 1:
+            quals["described_at_url"] = urls[0].strip_code()
+
+        date_templates = utils.wparser.parse(self.date).filter_templates()
```
lokal-profil: is it always the same template?`wondering if we might want to check that the template is the expected one.

importer/TnFr.py
```diff
@@ -0,0 +1,138 @@
...
+        if len(date_templates) == 1:
+            templ = date_templates[0].params
+            date_words = "{} {} {}".format(templ[0], templ[1], templ[2])
+            date_parsed = dateparser.parse(date_words)
```
lokal-profil: maybe add `languages=['fr']` for safety and paranoia

importer/TnFr.py
```diff
@@ -0,0 +1,138 @@
...
+    def set_heritage_id(self):
+        wlm_code = self.mapping["table_name"].upper()
+        wlm_id = "{}-{}".format(wlm_code, str(self.id))
+        self.add_statement("wlm_id", wlm_id)
+
+    def update_descriptions(self):
+        english = "listed monument of Tunisia"
+        self.add_description("en", english)
+
+    def exists_with_monument_article(self, language):
```
lokal-profil: Think this can be dropped (standard field name and mapping language)

importer/TnFr.py
lokal-profil: Done

importer/TnFr.py
lokal-profil: added check for correct template

importer/TnFr.py
lokal-profil: Done (but changed to >0 to also catch the incorrect template case

importer/TnFr.py
lokal-profil: added

importer/TnFr.py
lokal-profil: Done

tests/test_importer_utils.py
lokal-profil: Done

PR #104, opened 2017-09-13
Philippines

Vesihiisi:


importer/PhEn.py
```diff
@@ -0,0 +1,102 @@
...
+                                                        search_in="iso")
+            if len(iso_match) == 1:
+                adm_q = iso_match[0]
+
+        if adm_q:
+            self.add_statement("located_adm", adm_q)
+        else:
+            self.add_to_report("province", self.province, "located_adm")
+
+    def set_wd_item_from_name(self):
```
lokal-profil: It might be safer to here do:
```python
if utils.count_wikilinks(self.site_name) == 1:
    wikilink = utils.get_wikilinks(text)[0]
    self.monument_article = wikilink.title
```
That way you may then reuse the normal logic of `exists_with_monument_article()` (including section handling and blacklist integration)

importer/PhEn.py
```diff
@@ -0,0 +1,102 @@
...
+        self.set_monuments_all_id("cp_wmph_id")
+        self.set_registrant_url()
+        self.set_changed()
+        self.set_wlm_source()
+        self.set_country()
+        self.set_adm_location()
+        self.set_location()
+        self.set_heritage()
+        self.set_heritage_id()
+        self.set_coords()
```
lokal-profil: See https://phabricator.wikimedia.org/T176845 for coordinates issue. Essentially we'll have to build an own mechanism for this.

PR #76, opened 2017-08-17
Colombia

Vesihiisi:
Base class for Colombia.

Mapping: https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/co_(es)
Preview: https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/co_(es)/preview
Task: https://phabricator.wikimedia.org/T173929

PR #105, opened 2017-09-13
United States

Vesihiisi:


importer/UsEn.py
```diff
@@ -0,0 +1,110 @@
...
+
+        self.add_statement("heritage_status", her_q, quals=qualifier)
+
+    def update_labels(self):
+        self.add_label("en", utils.remove_markup(self.name))
+
+    def set_heritage_id(self):
+        self.add_statement("nhrp", str(self.refnum))
+
+    def exists_with_monument_article(self, language):
```
lokal-profil: Similar to the South American datasets the US dataset either uses `article` for an explicit link or generates a wiki link like `[[name]]`.

In the case where `article` is blank it might therefore be worth checking if name is an existing article.

importer/UsEn.py
```diff
@@ -0,0 +1,110 @@
...
+        against external list.
+        """
+        adm_q = None
+        county_dic = self.data_files["counties"]
+        state_dic = self.data_files["states"]
+
+        if utils.count_wikilinks(self.county) == 1:
+            county_q = utils.q_from_first_wikilink("en", self.county)
+            if county_q in [x["item"] for x in county_dic]:
+                adm_q = county_q
```
lokal-profil: add to report already here. We care about a failed match even if we got a conciliatory state match

importer/UsEn.py
```diff
@@ -0,0 +1,110 @@
...
+        if len(her_match) == 1:
+            her_q = her_match[0]
+
+        self.add_statement("heritage_status", her_q, quals=qualifier)
+
+    def update_labels(self):
+        self.add_label("en", utils.remove_markup(self.name))
+
+    def set_heritage_id(self):
+        self.add_statement("nhrp", str(self.refnum))
```
lokal-profil: Per [Template:NRHP_row](https://en.wikipedia.org/wiki/Template:NRHP_row) this value might contain a list of ids (comma separated or as a `<!-- comment -->`). We only need the first (per the same documentation) but should make sure that one is isolated.

importer/UsEn.py
```diff
@@ -0,0 +1,110 @@
...
+        If that doesn't yield, use ISO code of State
+        against external list.
+        """
+        adm_q = None
+        county_dic = self.data_files["counties"]
+        state_dic = self.data_files["states"]
+
+        if utils.count_wikilinks(self.county) == 1:
+            county_q = utils.q_from_first_wikilink("en", self.county)
+            if county_q in [x["item"] for x in county_dic]:
```
lokal-profil: use `utils.get_item_from_dict_by_key(county_dict, county_q, "item")`

importer/UsEn.py
```diff
@@ -0,0 +1,110 @@
...
+            if d_parsed:
+                date_dict = utils.datetime_object_to_dict(d_parsed)
+                qualifier = {"start_time": {"time_value": date_dict}}
+
+        her_raw = self.type.lower()
+
+        her_dic = self.data_files["her_types"]["mappings"]
+        her_match = utils.get_matching_items_from_dict(her_raw, her_dic)
+        if len(her_match) == 1:
+            her_q = her_match[0]
```
lokal-profil: else add to report

importer/UsEn.py
```diff
@@ -0,0 +1,110 @@
...
+        https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/us_(en)/type
+        If possible, add date of protection as quantifier.
+        """
+        qualifier = None
+
+        if self.has_non_empty_attribute("date"):
+            d_parsed = dateparser.parse(self.date)
+            if d_parsed:
+                date_dict = utils.datetime_object_to_dict(d_parsed)
+                qualifier = {"start_time": {"time_value": date_dict}}
```
lokal-profil: else add to report

importer/UsEn.py
lokal-profil: See https://github.com/Vesihiisi/COH-tools/pull/76#discussion_r136988915 for the coordinating comment around this.

PR #109, opened 2017-09-15
Armenia

Vesihiisi:
Mapping: https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/am_(hy)
Preview: https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/am_(hy)/preview

Task: https://phabricator.wikimedia.org/T176710


importer/AmHy.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+        if "monuments" in self.commonscat.lower():
+            return
+        else:
+            super().set_commonscat()
+
+    def set_heritage_id(self):
+        self.add_statement("armenia_heritage", str(self.id))
+
+    def exists_with_monument_article(self, language):
```
lokal-profil: can be dropped

lokal-profil:
A follow up on the `registrant_url` comment. Looks like there is an issue with the harvesting [T176991](https://phabricator.wikimedia.org/T176991) so it might have to be constructed manually here.

importer/AmHy.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def update_labels(self):
+        name = utils.remove_markup(self.description)
+        self.add_label("hy", name)
+
+    def update_descriptions(self):
+        english = "cultural heritage monument of Armenia"
+        self.add_description("en", english)
+
+    def set_adm_location(self):
```
lokal-profil: So I just spotted that the list also includes monuments in Nagorno-Karabakh (officially Azerbajdzjan). We therefore need to include a data_file for those communities and possibly manually create a data file for those regions (they have non-ISO iso codes which start with NK).

Additionally we need to set country to Q227 (Azerbajdzjan) for these.

Vesihiisi:
Added 'described at url' to heritage id and tested in sandbox.

importer/AmHy.py
Vesihiisi: Done.

importer/AmHy.py
Vesihiisi: Note: All the possible prov_iso are [here](https://gist.github.com/Vesihiisi/ffdbdcdd36abdcc012aa5704c8143399). So making a manual file for the NK ones

Vesihiisi:
Preview up to here.

PR #116, opened 2017-09-21
Thailand

Vesihiisi:
Preview: https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/th_(th)/preview
Mapping: https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/th_(th)

Task: https://phabricator.wikimedia.org/T176712


importer/ThTh.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def update_labels(self):
+        th = utils.remove_markup(self.name)
+        self.add_label("th", th)
+
+    def update_descriptions(self):
+        en = "cultural heritage site in Thailand"
+        self.add_description("en", en)
+
+    def set_wd_from_name(self):
```
lokal-profil: I've corrected the harvesting so that these should now show up in `monument_article` after the next harvest.

importer/ThTh.py
```diff
@@ -0,0 +1,83 @@
...
+        prov_dic = self.data_files["provinces"]
+        distr_dic = self.data_files["districts"]
+        iso = self.prov_iso
+        district = self.district
+
+        dist_match = utils.get_item_from_dict_by_key(dict_name=distr_dic,
+                                                     search_term=district,
+                                                     search_in="itemLabel")
+        if len(dist_match) == 1:
+            adm_q = dist_match[0]
```
lokal-profil: add to report already here. We care about a failed match even if we get a province later.

importer/ThTh.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def set_heritage_id(self):
+        wlm = self.mapping["table_name"].upper()
+        self.add_statement("wlm_id", "{}-{}".format(wlm, self.register))
+
+    def update_labels(self):
+        th = utils.remove_markup(self.name)
+        self.add_label("th", th)
+
+    def update_descriptions(self):
```
lokal-profil: You are not making use of this in `__init__`

lokal-profil (changes requested):
I corrected the harvesting so that it now gets `monument_article` and `registrant_url`.

The latter all seem to be broken though so they should be safe to skip (I left a message on the template discussion page in case the url just needs to be updated).

importer/ThTh.py
lokal-profil: Done

importer/ThTh.py
```diff
@@ -0,0 +1,112 @@
...
+                return super().set_heritage()
+        else:
+            return super().set_heritage()
+
+    def set_heritage_id(self):
+        wlm = self.mapping["table_name"].upper()
+        self.add_statement("wlm_id", "{}-{}".format(wlm, self.register))
+
+    def update_labels(self):
+        th = utils.remove_markup(self.name)
```
lokal-profil: I think we might also want to remove the stuff in brackets at the end?

lokal-profil:
Question about urls (and if id's might then actually be official) at [th:คุยเรื่องแม่แบบ:แถวโบราณสถาน](https://th.wikipedia.org/wiki/%E0%B8%84%E0%B8%B8%E0%B8%A2%E0%B9%80%E0%B8%A3%E0%B8%B7%E0%B9%88%E0%B8%AD%E0%B8%87%E0%B9%81%E0%B8%A1%E0%B9%88%E0%B9%81%E0%B8%9A%E0%B8%9A:%E0%B9%81%E0%B8%96%E0%B8%A7%E0%B9%82%E0%B8%9A%E0%B8%A3%E0%B8%B2%E0%B8%93%E0%B8%AA%E0%B8%96%E0%B8%B2%E0%B8%99#Broken_urls)

lokal-profil (changes requested):

importer/ThTh.py
lokal-profil: After the new harvest I believe this can be replaced by just `self.exists_with_monument_article(<lang code>)` since it's the default keyword.

lokal-profil:
Se changes to mapping per 
[Wikidata:WikiProject WLM/Mapping tables/th (th)](https://www.wikidata.org/w/index.php?title=Wikidata%3AWikiProject_WLM%2FMapping_tables%2Fth_%28th%29&type=revision&diff=588182680&oldid=568712762)

PR #120, opened 2017-09-28
Support monolingual text

lokal-profil:
Also introduce helper functions for packaging time, quantity and
monolingual text.


importer/Uploader.py
```diff
@@ -59,6 +59,15 @@ def make_quantity_item(self, quantity, repo):
             unit = None
         return pywikibot.WbQuantity(amount=value, unit=unit, site=repo)
 
+    def make_monolingual_item(self, quantity, repo):
+        """
+        Create claim for a monolingual text.
+
+        quantity: {'text': 'Q11573', 'lang': 6.85}
```
Vesihiisi: Maybe rename the variable and rewrite the example with actually expected values (strings, not qid and float)? I guess it's copypasted from make_quantity_item, took me a moment to parse.

importer/importer_utils.py
```diff
@@ -626,4 +626,21 @@ def create_site_instance(language, family):
...
+
+def package_quantity(value, unit=None):
+    """Package a quantity value in a standardised form."""
+    quantity = {"quantity_value": value}
+    if unit:
+        quantity["unit"] = unit
+    return quantity
+
+
+def package_time(text, date_dict):
```
Vesihiisi: The method doesn't make use of the `text` parameter.

importer/Uploader.py
lokal-profil: Yupp a whole lot of copypasta

importer/importer_utils.py
lokal-profil: good catch

PR #122, opened 2017-09-28
Only add default P31 if there is no previous P31

lokal-profil:
Adds a mechanism for only adding a claim if there are no previous
claims for this property.

Implements this mechanism for P31 when this takes the dfedault value.

Moves the de facto default P31 to common_items.json meaning it can
be removed from individual country mappings (but a country can use
a different default P31).

Task: [T176960](https://phabricator.wikimedia.org/T176960)


lokal-profil:
So with this any default P31 in the country mapping will always be added but the default one from common_items is only added if there is no P31.

I'm unsure whether country specific default one should also be made `if_empty=True`

importer/Uploader.py
```diff
@@ -159,6 +159,10 @@ def add_claims(self, wd_item, claims, log):
             for claim in claims:
                 prop = claim
                 for x in claims[claim]:
+                    if (x.get("if_empty") and
+                            self.item_has_prop(x["prop_name"], wd_item)):
```
Vesihiisi: You need to pass the name of the property here, so this won't work (because `x["prop_name"]` gives a KeyError).

A way to fix this might be modifying `item_has_prop()` so that it can take either a P-id or a property name, and then you could use `self.item_has_prop(prop, wd_item)`.

Vesihiisi: That is, if I understand correctly what you expected `x["prop_name"]` to return.

lokal-profil: So does `prop` and `claims[prop]["prop_name"]` both contain Pid?

Vesihiisi: `prop` is a Pid.

`claims[prop]` looks like this:

`[{'quals': {}, 'refs': [{'source': {'value': 'Q28563569', 'prop': 'P248'}, 'reference_url': {'value': 'https://tools.wmflabs.org/heritage/api/api.php?action=search&format=json&srcountry=mt&srlanguage=de&srid=146', 'prop': 'P854'}, 'published': {'value': {'month': 9, 'day': 14, 'year': 2017}, 'prop': 'P577'}}], 'value': 'Q19685249', 'if_empty': False}]`

lokal-profil: Thanks. new patch up.

And will rebase

lokal-profil:
Rebased. You may have to reload the page to make github not cry

PR #124, opened 2017-09-29
Change matched_item output from bulletpoints to wikitable

lokal-profil:
This will allow for easier viewing and grouping of `P31` types.

Task: [T176950](https://phabricator.wikimedia.org/T176950)


lokal-profil:
The output should en up like [this](https://www.wikidata.org/wiki/User:Andr%C3%A9_Costa_(WMSE)/test)

PR #126, opened 2017-09-29
Add method for dealing with duplicate ways of getting monument_article

lokal-profil:
This is the case for several South American datasets as well as us_en.
Coordinated through https://github.com/Vesihiisi/COH-tools/pull/76#discussion_r136988915

The way to implement it for e.g. Colombia is:
```python
def exists_with_monument_article(self, language):
    return get_multi_param_monument_article(
        raw="monumento_enlace",
        linked="monumento",
        blocks="enlace",
        language=language)
```


lokal-profil:
I added this as a PR separate from the affected countries to allow for an easier rebase.

PR #127, opened 2017-09-29
Ensure there is only one image in use

lokal-profil:
Wikidata recommendation is to only use one value for P18 so only
add if empty.


lokal-profil:
That it does =)

PR #133, opened 2017-10-06
Make utf-8 encoding explicit

lokal-profil:
While implicitly assuming this has worked fine so far (including
on Toolforge). However when running the command through jsub the
encoding needed to become explicit.

Task: [T177578](https://phabricator.wikimedia.org/T177578)


lokal-profil:
Not ready for review yet. awaiting a successfull run to see if anything else is needed.

lokal-profil:
Run was successful

PR #134, opened 2017-10-06
More compact matches_preview

lokal-profil:
The previous format produced tables which quickly surpassed the
allowed number of Lua calls on a page. This solution is more
resource demanding during a run but should produce a less demanding
output.


lokal-profil:
Not ready for review yet. awaiting a successful run to see if anything else is needed.

lokal-profil:
The new format can be seen at [Mapping_tables/ir_(fa)/matches](https://www.wikidata.org/wiki/Wikidata:WikiProject_WLM/Mapping_tables/ir_(fa)/matches)

Vesihiisi:
Is this ready for review then? Since the output looks super nice :D

lokal-profil:
Yes, sorry that I forgot to tag you

PR #136, opened 2017-10-17
[grid]Explicit encoding on opening file

lokal-profil:
Explicit encoding on opening file

For running on the grid the encoding must be explicit.

Also:
* Line breaking Readme

Task: [T177578](https://phabricator.wikimedia.org/T177578)


lokal-profil:
This is a follow up to  #133 which only implemented it on the bits used for creating the "matches" table

lokal-profil:
Comment on the README.

I discovered that running `--upload` throws an error (you cannot run it without a value). Running `--upload sandbox` gives the desired result though.

Note that setting `default="sandbox"` means upload is always detected as being present so that is also not desirable.

PR #137, opened 2017-10-30
[be_bru] Drop default P31 and remove 0 from street number

lokal-profil:
Per feedback in [bot request 8](https://www.wikidata.org/wiki/Wikidata:Requests_for_permissions/Bot/AliciaFagervingWMSE-bot_8)


importer/BeBruNl.py
```diff
@@ -47,7 +47,13 @@ def set_heritage(self):
...
     def set_address(self):
         street_with_no = utils.remove_markup(self.adres.title()).rstrip(',')
-        whole_address = "{}, {}".format(street_with_no, self.plaats)
+
+        # strip "0" as the number
+        streets = [street.partition(' 0')[0]
+                   for street in street_with_no.split(', ')]
+        street_with_non_zero_no = ', '.join(streets)
+
+        whole_address = "{}, {}".format(street_with_non_zero_no, self.plaats)
```
Vesihiisi: I wonder about situations with multiple streets (with zeros that we remove, indicating no exact address), e.g. https://tools.wmflabs.org/heritage/api/api.php?action=search&format=json&srcountry=be-bru&srid=2003-0021/0

This is some sort of park, other items with similar situations are of the type 'Halfnatuurlijk landschap' (some sort of nature area) etc. Does it really make sense to have a 'street address' property for those, if it ends up looking like "$street1, $street2, $street3, $town"? Since it's not really an 'address' per se.

lokal-profil: The multi street addresses are a bit weird (with or without street numbers). Since they are always streets at least the `located_street` Property should be ok but I'm open to suggestions.

Vesihiisi: I think I'm mildly okay with that, depending on how seriously we take the constraints on this property.

https://www.wikidata.org/wiki/Property_talk:P969

Looking at the P31 constraints - must be (subclass) of architectural structure etc... - which will fail on parks / nature areas.

lokal-profil: That one seems to be fairly frequently violated and at least parks can have addresses in many countries.

lokal-profil:
Updated preview

PR #138, opened 2017-10-30
Support pywikibot args

lokal-profil:
As these are always prefixed with a single "-" they should not
interfere with existing arguments.

The main use for this is being able to override which bot-account
is used for an upload. This can for some explicable reason not be
done using just `-user` but it's possible to do using `-dir` and a separate
user-config for that user. Leaving out the `-dir` runs the script using the default user.

Task: [T179275](https://phabricator.wikimedia.org/T179275)


lokal-profil:
Have now checked that this works correctly also when started using jsub

PR #139, opened 2017-10-30
Add a mechanism by which a disambiguator is registered and used

lokal-profil:
The disambiguator is added to the description if there is a need
to make it unique. Otherwise it is left out.

Task: [T179292](https://phabricator.wikimedia.org/T179292)


importer/Uploader.py
```diff
@@ -21,13 +21,56 @@ def add_labels(self, target_item, labels, log):
...
+                message = "{} ADDED DESCRIPTIONS {}".format(t_id, descriptions)
+                log.logit(message)
+
+    def add_disambiguators_to_descriptions(self, descriptions, api_error,
+                                           disambiguated):
+        """
+        Add needed disambiguators to descriptions.
+
+        :param descriptions: pre-existing descriptions
+        :param api_error: the raised pywikibot.data.api.APIError
```
Vesihiisi: Can be updated now that it works since we're not passing an APIError directly here.

lokal-profil: Done

PR #141, opened 2017-10-31
Strip() string uploads + BeBruNl street addr w/o plaats

Vesihiisi:
The bot crashed when trying to upload https://tools.wmflabs.org/heritage/api/api.php?action=search&format=json&srcountry=be-bru&srid=2076-0021/0 as it attempted to add the malformed input `Jachtlaan 141,[space that doesn't render here]`.

* Don't try to add self.plaats to street address if it's empty
* In general, strip() string values of claims before uploading to avoid trailing whitespace.

Task: https://phabricator.wikimedia.org/T179364


Vesihiisi:
Tested in https://www.wikidata.org/w/index.php?title=Q4115189&type=revision&diff=586704289&oldid=586704284, attempting to add `    whole_address 1    `

importer/Uploader.py
```diff
@@ -183,7 +183,7 @@ def make_pywikibot_item(self, value, prop=None):
         elif prop == PROPS["commonscat"] and utils.commonscat_exists(value):
             val_item = value
         else:
-            val_item = value
+            val_item = value.strip()
```
lokal-profil: Also add the strip() to L114 of Monuments.py (to make preview output the same as Uploader would)

Vesihiisi: Done.

PR #142, opened 2017-10-31
Sort matches by frequency

lokal-profil:
Apart from resulting ina amore sensible initial table it also reduces
the diffs between versions.

Task: [T179308](https://phabricator.wikimedia.org/T179308)


importer/importer.py
```diff
@@ -259,7 +259,10 @@ def get_items(connection,
 def format_matched_p31s_rows(matched_item_p31s):
     row = "|-\n| {p31} || {freq} || {hits} "
     rows = []
-    for p31, hits in matched_item_p31s.items():
+    sorted_items = sorted(matched_item_p31s.items(), key=lambda x: x[1],
```
Vesihiisi: It [works](https://www.wikidata.org/w/index.php?title=User:Alicia_Fagerving_(WMSE)/test&oldid=586759600) with `key=lambda x: len(x[1])`

PR #143, opened 2017-10-31
[Uploader/Monument] Handle None values

Vesihiisi:
Sometimes a `value` that is still None makes its way to Uploader / Preview, causing crash when trying to manipulate it...


importer/Uploader.py
```diff
@@ -183,7 +183,7 @@ def make_pywikibot_item(self, value, prop=None):
         elif prop == PROPS["commonscat"] and utils.commonscat_exists(value):
             val_item = value
         else:
-            val_item = value.strip()
+            val_item = value.strip() if value else None
```
lokal-profil: Also add something to the table preview?

Vesihiisi: Like just an empty row with the property name?

lokal-profil: Don't we today end up in `value_to_print += str(value).strip()` which would give `'None'?

lokal-profil: If so I would say we use `value_to_print += str(value).strip() if value else ""` or maybe just change the `else` to `elif value`

Vesihiisi: I thought with d0c9795b1e76ba4c9567c1a62b4c4e84a1c8fdbc we weouldn't get to `value_to_print += str(value).strip()` since it bypasses the rest of the block?

lokal-profil: I think I might be going blind...

PR #146, opened 2017-10-31
Allow the table preparation code to run on labs

lokal-profil:
Add entry points for labs database and support for pywikibot args.

Needed to set up mapping table for Albania as part of [T176118](https://phabricator.wikimedia.org/T176118)


lokal-profil:
I've tried it on labs and could generate a mapping table using `create_distinct_lookup_table.py --column type --table "monuments_al_(sq)" -dir:...`

PR #148, opened 2017-10-31
[GhEn] Use region iso instead of relying on wikilink

lokal-profil:
This was mentioned in #88 but missed in the followup

Task: [T174441](https://phabricator.wikimedia.org/T174441)


lokal-profil:
Now tested on labs and used to generate [a new preview](https://www.wikidata.org/w/index.php?title=Wikidata:WikiProject_WLM/Mapping_tables/gh_(en)/preview&oldid=587008020)

PR #149, opened 2017-11-01
[EsEs] Extend mappings by cases discovered in P31 matches

lokal-profil:
Task: [T179403](https://phabricator.wikimedia.org/T179403)


lokal-profil:
Some of the hits in matches today were already blacklisted in #145

lokal-profil:
The lake and watercourse were actually not wrong (surprisingly there are protected lakes and canals)

Q43229 is already in the blacklist (the matches table was generated using the pre #145 blacklist). Adding the rest.

Vesihiisi:
I started writing a long essay on the ontology of protected areas but seriously it sounded super petty so i deleted it.

PR #153, opened 2017-11-02
[co-es] Extend blacklist per Colombia matches

lokal-profil:
https://www.wikidata.org/w/index.php?title=Wikidata:WikiProject_WLM/Mapping_tables/co_(es)/matches&oldid=587401713

Task: [T179490](https://phabricator.wikimedia.org/T179490)


lokal-profil:
Q4671277 gets added din #154

lokal-profil:
Yes. But when I checked the actual objects that was what they said was protected...

Vesihiisi:
The problem as I see it is that our blacklist affects all datasets. So even if it was correct in this particular case (which it isn't -- the wikipedia modelling mixes up two things), it might cause problems further down the road.

Vesihiisi:
And since we already ban academic institution and the like, there's really no difference with faculty.

Vesihiisi:
> But when I checked the actual objects that was what they said was protected...

Nope, they just [included a link to the monument list in an awkward place and without explanation in an article that is 100% about the institution and 0% about the building(s)](https://es.wikipedia.org/wiki/Facultad_de_Medicina_Veterinaria_y_Zootecnia_(UNAL_Bogot%C3%A1))

PR #155, opened 2017-11-02
[util] Don't crash on missing page title

lokal-profil:
Because people do bad wikitext every day =(

Task: [T179566](https://phabricator.wikimedia.org/T179566)


Vesihiisi:
why would anyone do this

PR #162, opened 2017-11-03
[TnFr] Handle too long labels

lokal-profil:
This is unlikely to appear anywhere else. if it does then move
implementation to Monument.add_label()

Task: [T179725](https://phabricator.wikimedia.org/T179725)


Vesihiisi:
Is this ready for review?

Would there be a point of adding this as a general method to Monument.py?

Vesihiisi:
Is this ready for review?

I'm not that sure about the 'unlikely' part, though I can't give any examples. Feels like I have seen some monuments that use a free text 'description' field as label... :/

Vesihiisi:
With that I just mean it probably doesn't hurt to implement it directly in Monument.py since there's no overhead and we never know when it'll come in handy.

importer/Monument.py
```diff
@@ -240,6 +241,9 @@ def add_label(self, language, text):
         :param language: code of language, e.g. "fi"
         :param text: content of the label
         """
+        if len(text) >= MAX_LABEL_LENGTH:
+            self.add_to_report('', text, '_label')
```
Vesihiisi: So because of what the reporting system does with the third parameter (treats it like name of property and looks it up in the prop mapping file), this will look like:

```
    {
        "": {
            "target": null,
            "value": "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Maecenas neque augue, laoreet et dapibus ut, sodales in justo. Mauris pretium eros et purus pretium, non ultrices lorem tristique. Nulla eget semper magna. Duis mattis eget nibh vitae ultrices. Phasellus aliquet sem neque. Nunc ac urna dictum, eleifend mauris ut, tristique ipsum. Nulla metus."
        },
        "Q": "Q4115189",
        "url": "https://tools.wmflabs.org/heritage/api/api.php?action=search&format=json&srcountry=gh&srlang=en&srid=GH-CP-010"
    }
```

I.e. the `_label` never makes it here (target null). Maybe add it to the props_general file just to hack around it?

lokal-profil: Were does it do the lookup before dumpning to file?

Vesihiisi: looks like there's two relevant places. First I modified L245:

```
self.add_to_report('LABEL!', text, '_label')
```

Then added to props_general:
```
  "_label": "!LABEL"
```

Which resulted in:
```
    {
        "LABEL!": {
            "target": "!LABEL",
            "value": "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Maecenas neque augue, laoreet et dapibus ut, sodales in justo. Mauris pretium eros et purus pretium, non ultrices lorem tristique. Nulla eget semper magna. Duis mattis eget nibh vitae ultrices. Phasellus aliquet sem neque. Nunc ac urna dictum, eleifend mauris ut, tristique ipsum. Nulla metus."
        },
        "Q": "Q4115189",
        "url": "https://tools.wmflabs.org/heritage/api/api.php?action=search&format=json&srcountry=gh&srlang=en&srid=GH-TV-002"
    }
```

lokal-profil: Added the workaround in add_to_report. Might be useful for e.g. descriptions etc. later

PR #165, opened 2017-11-06
Bosnia

Vesihiisi:
Mapping: https://www.wikidata.org/w/index.php?title=User:Alicia_Fagerving_(WMSE)/Bosnia
Preview: https://www.wikidata.org/wiki/User:Alicia_Fagerving_(WMSE)/Bosnia_prev

Task: https://phabricator.wikimedia.org/T179828


lokal-profil:
Could you put together something like [https://www.wikidata.org/wiki/User:Eric_Luth_(WMSE)/Namibia](this matching table) to give an easy mapping overview from the source data columns to properties?

Vesihiisi:
Added a table here https://www.wikidata.org/w/index.php?title=User:Alicia_Fagerving_(WMSE)/Bosnia

importer/Bosnia.py
```diff
@@ -0,0 +1,321 @@
...
+
+        National Monument of Bosnia and Herzegovina (Q12637954)
+        with facultative start date.
+        """
+        heritage = self.mapping["heritage"]["item"]
+        date_parsed = dateparser.parse(self.designated)
+        quals = {}
+        if date_parsed:
+            date_dict = utils.datetime_object_to_dict(date_parsed)
+            quals["start_time"] = {"time_value": date_dict}
```
lokal-profil: utils.package_time(date_dict)

importer/Bosnia.py
```diff
@@ -0,0 +1,321 @@
...
+        type_match = utils.get_item_from_dict_by_key(
+            dict_name=type_dict,
+            search_in="itemLabel",
+            search_term=type_raw)
+        if len(type_match) == 1:
+            to_add = type_match[0]
+        else:
+            to_add = default
+
+        self.add_statement("is", to_add, refs=self.source)
```
lokal-profil: add `, if_empty=True)` to not add when there are pre-existing P31s

importer/Bosnia.py
```diff
@@ -0,0 +1,321 @@
...
+        if "N/A" not in coords_raw and "x" not in coords_raw:
+            if "N" in coords_raw and "E" in coords_raw:
+                split_parts = coords_raw.split("N")
+                try:
+                    coord_tuple = (float(split_parts[0].strip()),
+                                   float(split_parts[1].strip()[:-1].strip()))
+                    self.add_statement(
+                        "coordinates", coord_tuple, refs=self.source)
+                except ValueError:
+                    print("Nope coords: {}".format(self.coordinates))
```
lokal-profil: not add to report?

importer/Bosnia.py
```diff
@@ -0,0 +1,321 @@
...
+        self.set_coords()
+        self.set_is()
+        self.set_heritage()
+        self.set_described()
+        self.set_street_address()
+        self.update_descriptions()
+        self.update_labels()
+
+
+def generate_filenames(tablename, timestamp):
```
lokal-profil: why not just `import importer` in the start then call `importer.make_filenames` instead?

lokal-profil: True for several of the other functions as well and would make it easier to see which parts of the code are new and which are re-used.

importer/Bosnia.py
```diff
@@ -0,0 +1,321 @@
...
+        items[value] = key
+    print("FOUND {} WD ITEMS WITH PROP {}".format(len(items), prop))
+    return items
+
+
+def load_mapping_files():
+    """Load the files with mappings of various values."""
+    mapping_files = {}
+    mapping_files["_static"] = {}
+    files_to_get = ["ba_(bs).json", "bosnia_adm.json", "bosnia_types.json"]
```
lokal-profil: `ba_(bs).json` is missing from the patch

importer/Bosnia.py
```diff
@@ -0,0 +1,321 @@
...
+        self.existing = existing
+        self.repo = repository
+        self.problem_report = {}
+
+    def create_source(self):
+        source_item = "Q42750314"
+        last_edit = "2017-09-08"
+        timestamp = utils.date_to_dict(last_edit, "%Y-%m-%d")
+        prop_stated = self.props["stated_in"]
+        prop_date = self.props["publication_date"]
```
lokal-profil: Should this actually be "retrieval date" since the google doc isn't the actual date of publication?

lokal-profil:
Could we use the number in the url (just before the `?`) as an WLM identifier with a `BA-` prefix?

Vesihiisi:
> Could we use the number in the url (just before the `?`) as an WLM identifier with a `BA-` prefix?

Some of the items don't have url's (they haven't been uploaded to the system according to the notes). So the coverage of the WLM ID might be misleading as to the number of objects we've uploaded.

importer/Bosnia.py
Vesihiisi: :white_check_mark: this actually required hacking the Uploader.... :open_mouth:

Vesihiisi: ...though of course we don't _really_ know how accurate this date is, regardless of what prop we stick on it....

importer/Bosnia.py
Vesihiisi: :white_check_mark:

importer/Bosnia.py
Vesihiisi: :white_check_mark:

importer/Bosnia.py
Vesihiisi: :white_check_mark:

importer/Bosnia.py
Vesihiisi: Because at that point I had no way of knowing whether I'd be able to use the importer functions or whether and to what degree they'd have to be modified.

importer/Bosnia.py
Vesihiisi: :white_check_mark:

Vesihiisi:
preview up to here.

lokal-profil:
>> Could we use the number in the url (just before the `?`) as an WLM identifier with a `BA-` prefix?

> Some of the items don't have url's (they haven't been uploaded to the system according to the notes). So the coverage of the WLM ID might be misleading as to the number of objects we've uploaded. This also affects the finding of existing matches... There really isn't a good solution apart from not uploading the items without url's, which is not a good solution.

So what id are we adding today? Fro the preview it looks like we are not adding any, meaning we also cannot detect existing matches nor ensure that we don't upload the same item twice... Or am I misreading it?

lokal-profil:
Only looks like 6 out of the 170 don't have a url

Vesihiisi:
> So what id are we adding today? Fro the preview it looks like we are not adding any, meaning we also cannot detect existing matches nor ensure that we don't upload the same item twice... Or am I misreading it?

Yeah, there's no ID included right now and no, I don't like it either :/

Vesihiisi:
So I guess the extracted ID's from the URL's should be used as WLM ID, but what do with the non-urled ones?

lokal-profil:
>So I guess the extracted ID's from the URL's should be used as WLM ID, but what do with the non-urled ones?

Either skip them (for now) or assign them a differently structured identitfier (which we then also add to the document). E.g. `BA-1234` for url ones and `BA-wiki-01` for the others. Ideally we would propose a Property and then migrate the first type at some future point.

Vesihiisi:
Added WLM id from URL and skipped the ones without URL's... only 6 items as far as i can see.

Vesihiisi:
Preview updated here.

importer/Bosnia.py
```diff
@@ -197,6 +212,7 @@ def construct_wd_item(self, mapping, data_files=None):
         self.set_street_address()
         self.update_descriptions()
         self.update_labels()
+        # self.set_wd_item(self.find_matching_wikidata(mapping))
```
lokal-profil: Why is this commented out?

importer/Bosnia.py
```diff
@@ -98,20 +97,25 @@ def set_is(self):
...
         """
         if self.has_non_empty_attribute("url"):
             if self.url.startswith("http"):
-                url_clean = self.url.split("?")[0]
-                self.add_statement("described_at_url",
-                                   url_clean,
-                                   refs=self.source)
+                return self.url.split("?")[0]
+
+    def set_described(self):
```
lokal-profil: Do we actually want this as a qualifier on heritage status?

importer/Bosnia.py
Vesihiisi: ....because i forgot about it

importer/Bosnia.py
Vesihiisi: This is not being added as a qualifier on heritage status.

Vesihiisi: oh god now i get it

Vesihiisi:
Moved the url to quals of heritage status, updated preview, [pushed an item to the Sandbox](https://www.wikidata.org/w/index.php?title=Q4115189&oldid=590235063).

PR #167, opened 2017-11-06
[EsEs] Filter out non-BICs

lokal-profil:
The database holds lot of different id's many of which are not BIC

Task: [T179835](https://phabricator.wikimedia.org/T179835)


lokal-profil:
>> Can we spit out a list of all the items with bad ID's (within reasonable timeframe)?
We could run a pre-processing step but I'm not really sure it's worth it. The current solution should simply skip them.
I believe we could in theory upload some of the others but the easiest would be to do a separate class for those.

lokal-profil:
>Can we spit out a list of all the items with bad ID's (within reasonable timeframe)?

We could run a pre-processing step but I'm not really sure it's worth it. The current solution should simply skip them.
I believe we could in theory upload some of the others but the easiest would be to do a separate class for those.

PR #168, opened 2017-11-06
[IrFa] Strip preceding colons from filenames

lokal-profil:
Task: [T179858](https://phabricator.wikimedia.org/T179858)


lokal-profil:
As this is trivial I'l go ahead and merge it to give the upload more time to re-run

PR #169, opened 2017-11-06
[BoEs] Verify that adm links are known adm objects

lokal-profil:
No more adding [blanket](https://www.wikidata.org/wiki/Q5852) as an administrative region

Task: [T179559](https://phabricator.wikimedia.org/T179559)


lokal-profil:
I re-ran matches and preview to make sure this didn't cause anything unexpected

PR #171, opened 2017-11-09
[Uploader]Look for the upload flag in the right place

lokal-profil:


lokal-profil:
Just spotted one more issue. Not ready

importer/Monument.py
```diff
@@ -37,6 +37,7 @@ def __init__(self, db_row_dict, mapping, data_files, existing, repository):
         self.existing = existing
         self.repo = repository
         self.problem_report = {}
+        self.upload = True  # assume that upload of this monument is ok
```
Vesihiisi: Can this be somehow connected to the `self.wd_item["upload"]` in `construct_wd_item.()`? Because now it looks like they can have different values? Because it's unclear which one to use from the specific country class. Or maybe just remove the second one totally.

lokal-profil: I'm not entirely clear if there is a difference between the two. If not then I suggest dropping `self.wd_item["upload"]`.

Looking around I think Bosnia is the only one making use of `self.wd_item["upload"]`

lokal-profil: Ah. Because it mimics `Monument.py.construct_wd_item.()`

PR #172, opened 2017-11-09
[Importer]Output a list of skipped files

lokal-profil:
Allows re-run (without upload) of old runs to see any items
which should have been skipped but weren't.


importer/importer.py
```diff
@@ -260,6 +265,13 @@ def get_items(connection,
         print("\n")  # linebreak needed in case of visual feedback dots
     if problem_reports:
         print("SAVED PROBLEM REPORTS TO {}".format(filenames['reports']))
+    if skipped_uploads:
+        skipped_items_output += "\n".join(
```
Vesihiisi: `skipped_items_output` is never initialized so it will crash because referenced before assignment.

lokal-profil:
That square is a non rendering facepalm. Which pretty much sums up my handling of this patch =)

PR #174, opened 2017-11-09
[Uploader]Cannot have same function and var name

lokal-profil:
Scripts were not happy with both an `.upload` and a `.upload()`


Vesihiisi:
Well that's a stupid thing. Approved from my phone.

PR #175, opened 2017-11-11
Replace item redirect by target

lokal-profil:


lokal-profil:
Trivial so merging straight away

